PR #29, opened 2020-06-18
Upon changing/entering the Nightscout URL, strip any trailing spaces.

paulplant:
When the Nightscout URL is added/edited, some keyboards (SwiftKey in this case) will automatically add an "invisible" space to the end of the URL if the text is pulled from the auto-correct dictionary.

This will cause the API call to fail and not return data.

image


jonfawcett:
Nice catch.

PR #70, opened 2020-07-17
Merge All Latest Changes

josep1972:
This merges all jp changes with jf changes


jonfawcett:
Some of the new setting VCs run an immediate update of BGGraphSettings() or other items when changed. This previously just traversed the tabBarController to grab the appropriate VC to run the command in. It will need done with delegates now that we are 1 more level deep. You can search the new VCs for tabBarController! to find them all

jonfawcett:
Debug log causes similar issues with the way we destroy the tab bar item

josep1972:
I thing I fixed the settings. Not sure where to look for debug log, though.

PR #74, opened 2020-07-21
Merge latest

josep1972:
Implements authentication delegates infrastructure and implements dex delegate


josep1972:
Merge now has fix for mmol/L in std dev

jonfawcett:
After some testing, I don't think this is the way to go because it presents a lot of user disruption when there are failures.

PR #87, opened 2020-08-02
Output standard deviation in mmol/L when required

andrew-dixon:
Convert the standard deviation to mmol/L when output units are set to mmol/L in settings.


jonfawcett:
Sorry, I missed the PR until now. Please push all changes to Dev instead of Main. I'm going to close this so you can resend to Dev.

PR #93, opened 2020-09-14
Output standard deviation in mmol/L when required

andrew-dixon:
Convert the standard deviation to mmol/L when output units are set to mmol/L in settings.


andrew-dixon:
@jonfawcett As requested on #87

jonfawcett:
Merged

PR #101, opened 2020-10-18
Handling of null bolus

bjorkert:
If a bolus is cancelled it generates a bolus with a null value.

Current implementation in dev gives this error:
image

In this pull request I have a suggestion that works fine for me with a null-bolus.

### DIFF
--- a/LoopFollow/Controllers/NightScout.swift
+++ b/LoopFollow/Controllers/NightScout.swift
@@ -1337,8 +1337,9 @@ extension MainViewController {
             dateFormatter.timeZone = TimeZone(abbreviation: "UTC")
             let dateString = dateFormatter.date(from: strippedZone)
             let dateTimeStamp = dateString!.timeIntervalSince1970
-            do {
-                let bolus = try currentEntry?["insulin"] as! Double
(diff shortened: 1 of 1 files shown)


jonfawcett:
Thanks for helping. It looks like I messed up the original do-try-catch block based on the yellow warning notes. I thought that should have also caught null since that wouldn't be able to explicitly cast as! Double. Your code will solve null but fail on the other scenarios where it has a value but is not a number. For instance entering a bolus manually in care portal creates it as a string and not a number. I think maybe this should be a guard statement instead of do-try-catch. Can you check the below code run against your null value?

Remove the do and catch, leaving only the actual code
Replace the let bolus line with: guard let bolus = currentEntry?["insulin"] as? Double else { continue }

I *think* this guard statement will kick out if it cannot cast the value as Double which would cover both null and non-number values.

Screen Shot 2020-10-18 at 4 33 38 PM

bjorkert:
That code works fine! Will you implement it or should I replace it in this pull request?

jonfawcett:
I pushed it since it was just a 1 line change and I had some others to push in today. Thanks for your work catching this issue.

PR #106, opened 2020-10-29
Adding support for HbA1c when using mmol/L

bjorkert:
A1C to HbA1c formula from here: http://www.ngsp.org/ifccngsp.asp

Example:
image

### DIFF
--- a/LoopFollow/Application/Base.lproj/Main.storyboard
+++ b/LoopFollow/Application/Base.lproj/Main.storyboard
@@ -3,7 +3,7 @@
     <device id="retina6_5" orientation="portrait" appearance="light"/>
     <dependencies>
         <deployment identifier="iOS"/>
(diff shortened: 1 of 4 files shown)


jonfawcett:
This was already in but commented out (Line 75 in stats.swift). I took a rough poll early on and mmol users were split on which units they used for A1C calculations. Only a few of the mmol users used the IFCC format, so I had to make a decision one way or the other. But you bringing this up reminded me of it and here's the simple solution for how it should be implemented to allow for both options for mmol users.
1. add a user defaults variable useIFCC bool - default to false
2. add a toggle switch to general settings near the display stats one to edit the default: "Use IFCC A1C"
3. in stats.swift, uncomment line 75, remove line 72, change the if on line 69 to add: || !UserDefaultsRepository.useIFCC.value

If you want to edit the decimal display, use this same if statement like line 69 in StatsView.

PR #107, opened 2020-10-30
setting to use useIFCC A1C

bjorkert:
New setting for IFCC A1C
image

Default false

Toggling updates the stats view.

IFCC A1C can be used regardless of unit (mg/dl - mmol/L)


jonfawcett:
Nicely done.

Issue #105, opened 2020-10-27
"Keep Screen Active" reverts back ON if turned OFF


bjornoleh:
The General Settings - "Keep Screen Active" always seem to revert back ON if turned OFF
(dev as of today)

PR #109, opened 2020-11-07
Keep Screen Active reverts back ON if turned OFF #105

bjorkert:
Fix for issue #105


jonfawcett:
Thanks! No idea why I would have made that code so blatantly wrong 😂

PR #127, opened 2021-01-31
Added Repeat Alerts to Quiet Hours

mikebash:
Added an option to repeat alerts when quiet hours are on.  Forces all critical alerts to repeat regardless of the setting in the individual alert.  Idea is that you can have alerts not repeat during the day but at night when sleeping, the alerts do repeat to help wake someone up.  Alerts defined as critical are Urgent Low, Low, High, Urgent High, No Readings, No Looping, Fast Drop, and Fast Rise.  Note that the alerts will **not** fire (and therefore not repeat) if the alert is not active. This change only affects active alerts.

### DIFF
--- a/AlarmViewController.swift
+++ b/AlarmViewController.swift
@@ -1,0 +1,2230 @@
+//
+//  AlarmViewController.swift
+//  LoopFollow
+//
+//  Created by Jon Fawcett on 6/3/20.
+//  Copyright © 2020 Jon Fawcett. All rights reserved.
+//
+
+
+import UIKit
+import Eureka
+
+class AlarmViewController: FormViewController {
+    var appStateController: AppStateController?
+    
+    var soundFiles: [String] = [
+        "Alarm_Buzzer",
+        "Alarm_Clock",
+        "Alert_Tone_Busy",
+        "Alert_Tone_Ringtone_1",
+        "Alert_Tone_Ringtone_2",
+        "Alien_Siren",
+        "Ambulance",
+        "Analog_Watch_Alarm",
+        "Big_Clock_Ticking",
+        "Burglar_Alarm_Siren_1",
+        "Burglar_Alarm_Siren_2",
+        "Cartoon_Ascend_Climb_Sneaky",
+        "Cartoon_Ascend_Then_Descend",
+        "Cartoon_Bounce_To_Ceiling",
+        "Cartoon_Dreamy_Glissando_Harp",
+        "Cartoon_Fail_Strings_Trumpet",
+        "Cartoon_Machine_Clumsy_Loop",
+        "Cartoon_Siren",
+        "Cartoon_Tip_Toe_Sneaky_Walk",
+        "Cartoon_Uh_Oh",
+        "Cartoon_Villain_Horns",
+        "Cell_Phone_Ring_Tone",
+        "Chimes_Glassy",
+        "Computer_Magic",
+        "CSFX-2_Alarm",
+        "Cuckoo_Clock",
+        "Dhol_Shuffleloop",
+        "Discreet",
+        "Early_Sunrise",
+        "Emergency_Alarm_Carbon_Monoxide",
+        "Emergency_Alarm_Siren",
+        "Emergency_Alarm",
+        "Ending_Reached",
+        "Fly",
+        "Ghost_Hover",
+        "Good_Morning",
+        "Hell_Yeah_Somewhat_Calmer",
+        "In_A_Hurry",
+        "Indeed",
+        "Insistently",
+        "Jingle_All_The_Way",
+        "Laser_Shoot",
+        "Machine_Charge",
+        "Magical_Twinkle",
+        "Marching_Heavy_Footed_Fat_Elephants",
+        "Marimba_Descend",
+        "Marimba_Flutter_or_Shake",
+        "Martian_Gun",
+        "Martian_Scanner",
+        "Metallic",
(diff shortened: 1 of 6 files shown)


jonfawcett:
Thanks for tackling this! I haven't had time to build the PR yet, but skimmed the code and have a few suggestions:
- I think we should make a toggle 'Use Quiet Hour Repeat' under each of the alerts similar to how I did the quiet hour toggles. Not everyone will want the same alert list for the repeat. This would mean the toggle under the QH area is moved to one under each of these alerts.
- With this addition, I'm thinking we should maybe set default quiet hours rather than nil. With my original ones, it just meant alerts went off when you didn't want them to if you didn't set the times. With these, it means they won't repeat when you want them to. So I'm concerned if people don't set the time.
- Maybe Quiet Hours should be renamed to Nighttime Settings now? This will give us the start to add nighttime dedicated volume levels and other features.

mikebash:
Hi Jon - yes, all makes sense.  I created this trimmed down version first as I didn't know what others would think but I was also considering if there should be a repeat option under each alarm.  I'll spend some time this week attempting to make the updates in your first bullet.  I'm definitely supportive of changing Quiet Time to Nighttime as I expect that's really how people will use it (when they plan to be sleeping).

I'll submit a new PR once I have something worth looking at.  

Thanks for the suggestions!

mikebash:
@jonfawcett Do you think the Repeat Alerts selection should be a spinner?  I currently see 4 potential options:

- Do not repeat alerts
- Always repeat alerts
- Only repeat alerts during the night time
- Only repeat alerts during the day time

Each alarm could have these 4 options (more if there are any I'm not thinking of) and instead of multiple lines in the alarm settings, it would be a single line with these four options.

Thoughts?

jonfawcett:
Love this idea! How are you thinking of storing the names/values since it's no longer a simple bool like you had before? For instance, we obviously don't want a bunch of: if str = "Repeat only at night".

mikebash:
I'm still working that out in my head.  While there are four options, it really is just two bools (on/off in day and on/off at night).  So likely convert the four options into a bool for day and a bool for night.  But I'm open to other suggestions.

mikebash:
Hi @jonfawcett - I'm running into an error and not finding any good help when googling so I thought, perhaps, you might have a suggestion.  More details below but I'm trying to call a function where I pass the value of the picker and swift is telling me: 

`Call to method 'timeBasedSettings' in closure requires explicit use of 'self' to make capture semantics explicit`

I'm struggling to figure out how to solve this.  There's a lot of [good Stack Overflow conversations](https://stackoverflow.com/questions/32665326/reference-to-property-in-closure-requires-explicit-self-to-make-capture-seman) on this topic but I'm not seeing how to translate the examples into useable solutions.

Here's what I'm trying to do:

Given that every alarm will have the options for when to repeat, I made a function to translate the item selected in the picker into two bools that say if the alarm should repeat during the day and if it should repeat at night.

Here are the options for the picker
```
 var alertRepeatOptions: [String] = [
        "Do not repeat",
        "Always repeat",
        "Repeat at night",
        "Repeat during the day"
    ] 

```
And here is the function
```
func timeBasedSettings (pickerValue: String) -> (dayTime:Bool, nightTime:Bool) {
        var dayTime = false
        var nightTime = false
        
        if pickerValue.contains("Always") {
            dayTime = true
            nightTime = true
        }else{
            if pickerValue.contains("night"){
                nightTime = true
            }
            if pickerValue.contains("day"){
                dayTime = true
            }
        }
        return (dayTime, nightTime)
    }
```
Lastly, when I'm in the .onChange section I try to call the function `let alertTimes = timeBasedSettings(pickerValue: value)` and this is where I get the error above.

I understand conceptually why swift wants what it wants but I can't figure out how to use self and/or weak to accomplish this.  If you have any pointers on articles to read or solutions, I'd appreciate them.  Thanks.

jonfawcett:
You need to call it like this:
let alertTimes = self.timeBasedSettings(pickerValue: value)

mikebash:
Ha!  I feel like I tried every configuration except that one.  I've got to do a lot more reading because that's a topic I'm pretty lost on.  But THANK YOU!

Also, swift forces one of the following two options:

` let alertTimes = self!.timeBasedSettings(pickerValue: value)`

` let alertTimes = self?.timeBasedSettings(pickerValue: value)`

I think I'll use ! based on the description `Force-unwrap using '!' to abort execution if the optional value contains 'nil'`

jonfawcett:
In this code, ! Should be fine. If it could ever possibly be nil, then ! will crash the app if it returns nil (unless you use guard on the let). But your code looks like it will always return correctly.

mikebash:
I didn't realize it would crash, I thought it would simply not call the function.  Good to know.  And yes, since it should never be nil, shouldn't be an issue.

mikebash:
I'm having issues with github/xcode (sorry) but assuming I didn't completely botch it all, here's the updated code for changing quiet hours to night time and enabling different repeat alerts depending on the time of day.  Please take a look and let me know your feedback.  Thanks!

mikebash:
Actually, @jonfawcett, after thinking more there are some aspects of my changes that I want to update.  This is a move in the right direction but it's not ready to merge in.

jonfawcett:
Ok I’ll hold off on testing it yet

mikebash:
Given my GitHub issues and making a few additional changes, I've created a new pull request (#129) to start over instead of building on this one

PR #129, opened 2021-02-09
Extended repeat alert options

mikebash:
Gave all the alerts the option to:

- Never repeat
- Always repeat
- Repeat only at night
- Repeat only during the day

Changed "quiet time" to "night time" to simplify the concept and create future opportunities for further enhancements differentiating how the app responds in the day vs night.


mikebash:
@jonfawcett I'm trying to duplicate the options above for overriding the system volume so that you can choose different volume behaviors during the day and at night.  As I've worked on this, I believe that the current set up does not respect the mute switch.  Is that correct?  It seems that the sounds will play when muted as long as volume isn't 0 (or if you override it).  Curious if this is intentional and if so could you tell me a little more about your thinking on what you do / don't want for sounds?

My in-going thought was that during the day, when I'm wearing my apple watch, I want the option to silence my phone of all sounds (LoopFollow and others) via the mute switch.  Then at night, if I forget to unmute (or want to mute everything but LoopFollow) I'd want LoopFollow to bust through all sound settings.  So using the same approach I created for repeats, I've tried out implementing system override day vs night options.  This is where I discovered that I'm still getting sounds even when the mute switch is flipped.

Thanks for your input!  No rush.

jonfawcett:
Yes it’s intentional to break through mute and dnd. That’s the only way to guarantee you hear them if they fire at night. I had too many times in the past where I accidentally had my switch on vibrate going to bed and forgot. I think the easiest solution is to make the 2 trigger alarm methods have another bool for audio. Then in he methods, just a simple “if audio” around the couple of lines playing the sound. FYI, the timer in there is used to stop the audio and reset the snoozer screen at the next 5 min interval. This lets us turn off alarms if the situation passes and it hasn’t been snoozed. I think you’ll need to still start the timer, but add an if around the stop audio method to only stop if it’s playing. Not sure what will happen if trying to stop when not playing.

IMG_0710

mikebash:
Thanks for the nudge in the right direction.  I'll take a look at this and see what I come up with.

mikebash:
Hi @jonfawcett - Do you mind explaining your thoughts above about the audio bool in a little more depth?  I'm not totally following and as I've been exploring ways to make this work, I'm not finding an easy solution.  I'm probably just missing something.

My intended approach was to check if the mute switch on the side is on/off but it appears Apple doesn't make that easy.  So without being able to do that simply, how were you thinking about determining if LoopFollow should/shouldn't play audio?

jonfawcett:
My thought is not to change how LF treats the hardware switch or dnd, only when LF decides to play audio based on day/night. So trigger method would change to triggerAlarm(sound: String, SnoozedBGReadingTime: TimeInterval?, OverrideVolume: Bool, numLoops: Int, Audio: Bool = true) (set the default to true so we only have to deal with it when false).

I'm not sure what volume options you were considering, but this would allow sound always (bool = true 24x7), sound only during day (bool = false if night), sound only during night (bool = false if day)

mikebash:
Okay - I see where you are headed.  Ideally I'd like to figure out how to integrate the hardware switch but since Apple isn't making that easy, let's get a first version created that allows the user to select when the audio is played (or perhaps better said, when audio is muted).

With this in mind, I'll keep "Override System Volume" as a global setting (affects all alarms) but allow users to select for each individual alarm if it plays audio always, never, only during the day, or only at night.  This way, if the system volume override is on, when audio DOES play it will govern the volume.  When audio DOES NOT play it won't matter what the system override setting is.  If the user chooses to not turn on the system volume override, the alerts will always play at the volume the user has currently selected.

If you see anything above that's not lining up with your expectations, let me know.  Otherwise, I'll start implementing this and see how it works out.

jonfawcett:
This is seeming to become more complex than I understood and I'm starting to get confused as to what the purpose of some of these would be...which means users will probably become confused by complex settings. Why would you enable an alarm and have the audio set to never play? Wouldn't you just not enable that alarm? Can you also refresh my memory for the original idea you had on why you want an alarm to not play audio either during the day or during the night but still want the alarm to still trigger? The only thing that would happen is an iOS notification in that event. 

I'm just starting to wonder if we're making this too complex and will confuse people. So hopefully you can refresh my memory so we make the simplest way (and least confusing way) to make that function work.... less long-term code maintenance also

mikebash:
Hi @jonfawcett - Yes totally agree we don't want to confuse users.  Definitely good questions and worth stepping back to truly define the use case.

Simplifying this a little, there are times when a full audio alert is warranted and other times when a notification of sorts is needed but should be more discrete.  For instance, when sleeping, certain alerts need to wake me (highs, lows, fast drops, etc).  When I'm at work and on conference calls, I want to have notifications show up via the watch or phone (along with a vibration) but not make audible sounds that distract others.  This is what I'm trying to accomplish.  Giving the user the ability to silence sounds but still get notifications that something might need their attention. 

What are your thoughts on how to best implement this?  Totally open for discussion and happy to jump on the phone or a zoom call to debate the best approach if that's easier.

jonfawcett:
So I could see a few possible uses/needs for how it would work.

1. based on quiet hours like we discussed. I think this renames to 'night time hours' and then make toggles under the alarms. a) only iOS notification during day b) skip alert during night. c) skip alert during the day
There is 1 other possible options but I don't see a reason for it:
1. no sound during night (but still send the iOS notification). Since that won't wake you, just skip it during the night which snoozes it until the night hours end. So if the situation is still active (such as pump change) it then triggers immediately when those hours end.

2. maybe a much better solution. We could add a sound selection called "No Sound" at the top of the sound list and then edit the audio player to add at the beginning of the function: if sound == "No Sound" { return }. This will kick it out of the function and not play anything. Then we create 2 sound selections (Daytime Sound, Nighttime Sound). So for your case, you could pick No Sound for your daytime alarms and whatever sound you want for nighttime. Then we could still have the existing toggle to skip alerts during night but also add one to skip during day. So if someone wanted to rely only on dex alerts during day but LF alerts at night, they could do it. I think this gives us every possible combination in a very simple structure. Thinking a bit more now, maybe the enable line is just split and renamed. Enable during day, Enable during night. I'd have to think through that a little more.

3. One last option that would be the simplest is to add an additional snooze all function. "Mute Sound for All". So if you are walking into a meeting, you could trigger that for however long you want and it will still deliver quiet notifications. This one might be a really cool feature even if we add the more robust features.

mikebash:
Hi Jon - starting with #4, yes I was thinking about the exact same thing.  It would be nice to have an option to "mute all" in addition to "snooze all".  Regardless of everything else, I think that's worth implementing.

In my pull request above for the different repeat patterns, I did go ahead and change Quiet Hours to Night Time.  There's some clean up work still needed in the code (some of the variables still reference quiet hours) but it's now surfaced to users as Night Time if you were to integrate that pull request.

With regards to the alerts, I don't know that we would want to split the enable line (enable during the day, enable at night).  My gut says that people will decide that they want an alert but may want different behaviors from that alert depending on the situation.  So I lean towards having one enable button like you do now and then giving appropriate options for the alert's behaviors.

And I feel the same about daytime sound versus nighttime sound.  It's likely a better user experience to have one sound selector but give people the ability to snooze and/or mute sounds when they want to.

The more I think about the options we've both been putting out there, the more I go back to your original suggestion to manage the sound based on day versus night plus give people a "mute all" option.  This feels like simplest, most direct way to give people the full range of options without confusing users.  The alarm configuration window could look something like this:

image

Although I don't love the row title for when to play sounds ("Play alert sound").  Need to see if there's a more direct/clear way to articulate that.

Thanks for all your input on this.  Let me know what you think.

jonfawcett:
I’m good with this. So let’s do your picker with: Always, Only Night Time, Only Day Time. I think those will be most clear. And I think “Play Sound” is about as clear as anything and simple. Let’s default to Always. Should this be added to all alerts?

and then the other circumstances can all be handled with a new Mute All Sounds option next to snooze all.

mikebash:
On it!

PR #131, opened 2021-04-03
Nighttime vs Daytime controls

mikebash:
Created a structure to have night time settings that differ from day time (an extension of the previous Quiet Hours).  This code is still early in development and appears to be buggy.  Needs additional review and likely updates before it should be used.

image

For each of the alarms, new options are included such as when to play the sound (always, never, at night, during the day) and when to repeat.

image

There is also an option to mute all in addition to the existing snooze all.  Mute all stops the sound only (not the vibration or iOS notifications).  Snooze all continues to function as it did stopping all sounds/notifications for the specified period.

image

In theory, this should give the user more control on what sounds occur when.  In practice, it seems buggy, specifically sounds stop playing when they are supposed to.  I think what's happening is that after a sound is set to not play (e.g. during day time hours), that setting persists and the sound doesn't play during the evening/night when it's supposed to.  I need to read through the code again but there probably needs to be some sort of reset that occurs to clear out the variables.  Settings shouldn't carry over between daytime/nighttime, however, it appears they are.

All thoughts and suggestions are welcome.  Sharing this now as a work in progress to allows others to take a look if interested.  As stated above, this should not be used until further testing and updates are completed.


jonfawcett:
I built today and will test over the next few days. We'll try to address any immediate bugs (if any) and then merge to dev so more people can easily test

jonfawcett:
Here's a first few to get started. Not really bugs, but should be modified I think.
1. The old quiet hours would completely bypass an alert until the end time. The new method looks like it will still trigger the alert but just skip the sound. For SAGE and CAGE especially, that will have hundreds of notification bubbles when you awake since they will trigger every 5 minutes because the issue doesn't ever go away on its own like a BG alert. I have a 12 hour nighttime hours when I don't want the extraneous alerts, so on a day of dex and pod change I'd wake up to potentially 288 notification bubbles. I'm thinking it should be an easy modification to the existing checkQuietHours function. That essentially does a pre-snooze on the alert until the end time. So all we would need is a new setting in each of the alerts , maybe "Allow To Alert" - Always, At night, During the day (don't need the 'never' option on this one since that's voided by the alert's Active toggle).  How it currently works is checkQuietHours sees if the current time is within the range of nighttime. If it is, it calls setQuietHours in SnoozeViewController to set a pre-snooze until the end time. It might make more sense to add a new section inside the checkQuietHours to pull a list of all alarms with day only and night only. Then run that through a for loop based on day vs night to set the snoozes?

2. The 4 primary BG alerts should probably be set to Always play sound and Always repeat by default. And the rest probably set to Always play sound but never repeat by default? Thoughts?

jonfawcett:
Just had my first alert (fast rise) during the day with sound set to night only. My watched had haptic and displayed the notification. This is a brilliant feature for those less critical alerts 👍

jonfawcett:
@mikebash this is fabulous other than those 2 minor issues above. I ran yesterday with mute all day to test watch notifications and it’s working great. I think I’ll merge this into dev even without those changes so we can get more people testing

jonfawcett:
What do you think about a toggle on Urgent Low "Ignore Mute All"?

jonfawcett:
Bug: Not Looping alert triggered with no sound. It is currently “night” hours and it is set to play alert at night.

34BA9958-80D5-4873-9324-CA35297DF813
C5520CE3-A85C-4F59-9DB3-C685E9F00C3D

jonfawcett:
Just pushed the fix. The checkQuietHours function was only checking current time > nighttime start. This was originally like this because I used the start to check and then snooze the alerts until the end. So we didn't need to check the ending. So this meant only the pm start until midnight worked with the new code. Anything after midnight was no longer nighttime hours. I added an Or statement to catch the after midnight hours. So current minutes (0 = midnight to 1440 = midnight) > night start or less than night end so it catches everything between the 2.

mikebash:
Great!  Thanks @jonfawcett.  I had wondered a little about that when I was making the original changes but trusted it was there for a reason (and it was but clearly it was creating new problems based on my changes).  I suspect that's the major bug causing issues.  Let's see.  I'll rebuild with the new Dev version.

mikebash:
> What do you think about a toggle on Urgent Low "Ignore Mute All"?

Not a bad idea.  Should it only be this alarm? Probably but just thinking out loud.  And how should the Urgent Low react to the snooze all?  Still stay snoozed?

jonfawcett:
> Not a bad idea.  Should it only be this alarm? Probably but just thinking out loud.  And how should the Urgent Low react to the snooze all?  Still stay snoozed?

I think only that alarm. Low will turn into urgent low if it continues dropping and highs aren't the immediate life or death situation as urgent low. But if we're adding it for mute all, might as well add another toggle to ignore snooze all also.

mikebash:
Makes sense.  I'm not going to add these changes just yet as I'd like to see if there's any additional feedback from the community.  So probably give people a week or two just to see if people have other ideas, issues, etc.

jonfawcett:
Agreed. I only did that change today because it was major with alerts not sounding at night.

mikebash:
Yes - anything urgent I'm happy to jump into sooner.  Thanks for fixing the bug!

jonfawcett:
@mikebash I took care of this and realized how I did it introduces a new feature that some have asked for. I added "Pre-Snooze" to all alarms with Options: Never, At night, During day. Please build Dev and help me test these. Here is what this adds:
1. ability to auto snooze alerts at the start of night or day hours. This eliminates the issue of alerts like SAGE/CAGE that will continually trigger all night long even if there is no sound.
2. I've added this option to all alerts which allows people to use different devices at different times of the day without needing to force close the app or deal with alerts ringing from somewhere else all the time. iPad or old phone kept in the bedroom could be set to auto snooze during the day, Mac could be set to auto snooze outside work hours, etc.

> 1. The old quiet hours would completely bypass an alert until the end time. The new method looks like it will still trigger the alert but just skip the sound. For SAGE and CAGE especially, that will have hundreds of notification bubbles when you awake since they will trigger every 5 minutes because the issue doesn't ever go away on its own like a BG alert. I have a 12 hour nighttime hours when I don't want the extraneous alerts, so on a day of dex and pod change I'd wake up to potentially 288 notification bubbles. I'm thinking it should be an easy modification to the existing checkQuietHours function. That essentially does a pre-snooze on the alert until the end time. So all we would need is a new setting in each of the alerts , maybe "Allow To Alert" - Always, At night, During the day (don't need the 'never' option on this one since that's voided by the alert's Active toggle).  How it currently works is checkQuietHours sees if the current time is within the range of nighttime. If it is, it calls setQuietHours in SnoozeViewController to set a pre-snooze until the end time. It might make more sense to add a new section inside the checkQuietHours to pull a list of all alarms with day only and night only. Then run that through a for loop based on day vs night to set the snoozes?

mikebash:
Looks cool! Thanks for the heads up. I'll build Dev today or tomorrow and try it out.

PR #136, opened 2021-08-02
    Add code to allow reporting the following for *APS NS data:

scrappy:
- Basal Rate
      - IOB
      - COB

    Change label from Pump -> Reservoir to be clearer on what this is for

### DIFF
--- a/LoopFollow/Controllers/NightScout.swift
+++ b/LoopFollow/Controllers/NightScout.swift
@@ -459,7 +459,31 @@ extension MainViewController {
                 }
             }
         }
-        
+       
+        // OpenAPS ( aka FreeAPS X aka AndroidAPS )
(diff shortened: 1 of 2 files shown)


jonfawcett:
Thanks for starting this.  All PRs need done against the Dev branch so I’m going to close this so you can open it against the correct branch.

Please also find some testers of freeapsx and openaps that can test your PR code and comment on their after you open it. Unfortunately I have no way of testing it and can only test that if doesn’t mess up something for Loopers.

last thing, Please keep anything not related to freeaps out of the PR (eg the pump name change).

sethgagnon:
I tested this PR and it seems to be working. See attached screenshot which accurately shows current basal rate, IOB, and COB.  CAGE is accurate. SAGE is not showing. 

D3A6E7A8-6B70-452C-A46F-0F8DC1AE0338_1_105_c

I know this was closed, but maybe @scrappy can submit the PR against DEV branch using this as testing?

PR #138, opened 2021-09-07
Show more than 1 day of data in the chart

ostap-korkuna:
Adding an ability to show more than 1 day of data in the chart.

Nightscout API by default supports up to 4 days of history. If both Dexcom share and Nightscout are configured, first 24 hrs are loaded from Dexcom and the remainder from Nightscout.

Also adding a setting to show vertical lines at midnight: convenient when looking at multiple days of data.

### DIFF
--- a/LoopFollow/Controllers/AppStateController.swift
+++ b/LoopFollow/Controllers/AppStateController.swift
@@ -28,6 +28,8 @@ enum ChartSettingsChangeEnum: Int {
   case lowLineChanged = 512
   case highLineChanged = 1024
   case smallGraphHeight = 2048
+  case showDIALinesChanged = 4096
+  case showMidnightLinesChanged = 8192
 }
 
 // General Settings Flags
--- a/LoopFollow/Controllers/Graphs.swift
+++ b/LoopFollow/Controllers/Graphs.swift
@@ -305,6 +305,7 @@ extension MainViewController {
         
         // Add Now Line
         createNowAndDIALines()
+        createMidnightLines()
         startGraphNowTimer()
         
         // Setup the main graph overall details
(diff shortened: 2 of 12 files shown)


ostap-korkuna:
Now correctly merging Dexcom and Nightscout data.

Still need to fix the aggregated stats at the bottom to only aggregate 1 day of data instead of everything that's on the chart. One day is likely more useful for most people.

jonfawcett:
Building to test now.

jonfawcett:
Doesn't seem to have any major issues. A few minor tweaks that I'd like before merging though. Let me know your thoughts.

1. Dexcom/Nightscout text above the BG should not list both together. Please revert your change on this. That communicates the source of the current BG reading to the user.
2. On changes to your settings (midnight lines), update the AppState so it can trigger immediate redraws as needed. Search appState.chartSettingsChanged in GraphSettingsViewController (the onchange event of the setting) and MainViewController (the function that gets called). Then look at updateBGGraphSettings() in Graphs.swift. You'll probably need to add some code here. FYI I do not have anything to update the actual graph data as this proved problematic if making a bunch of quick changes that forced new data. So I think we just need to wait on the 5 min update for turning on the additional days.
3. Let's go ahead and put the # Days settings into the Graph Settings page rather than Advanced. I think for the next release I'm going to get rid of Advanced.
4. Can you add the Midnight Lines to the small graph? I think they will be even more beneficial in that one than in the larger graph. They might need to be more prominent of a line to see well that small though.

ostap-korkuna:
@jonfawcett good suggestions! I'll incorporate!

jonfawcett:
How’s it coming for #3-4? I’m getting close to being ready to push all recent changes to master as a major update and was hoping to include this.

ostap-korkuna:
Sorry, have been traveling and thus didn't get to it earlier. I only need to finish 4, everything else is done. Will try to find time today to wrap it up.

BTW, while traveling thought of another really useful feature: using time zone from NS. But that's for another release :)

ostap-korkuna:
@jonfawcett incorporated all your suggestions. 

Changed the midnight lines color a bit to make them more visible. Looks fine on my phone:
image

jonfawcett:
Testing

jonfawcett:
great new feature! Just merged it to Dev

ostap-korkuna:
Awesome! Thank you!

PR #143, opened 2021-10-12
fix bug where app crashes because of Could not cast value of type 'NS…

vallard:
My app kept crashing when it received data from NightScout when I compiled it and ran on my phone.  
I found I got the error: 
```
Could not cast value of type 'NSTaggedPointerString' to 'NSNumber'
```

I found a way to fix the error by using this suggestion from Stack Overflow: 
https://stackoverflow.com/questions/33299202/could-not-cast-value-of-type-nstaggedpointerstring-to-nsnumber

### DIFF
--- a/LoopFollow/Controllers/NightScout.swift
+++ b/LoopFollow/Controllers/NightScout.swift
@@ -837,9 +837,9 @@ extension MainViewController {
         for i in 0..<basal.count {
             let dict = basal[i] as! Dictionary<String, Any>
             do {
-                let thisValue = try dict[keyPath: "value"] as! Double
+                let thisValue = try (dict[keyPath: "value"] as! NSString).doubleValue
                 let thisTime = dict[keyPath: "time"] as! String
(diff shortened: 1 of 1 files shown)


jonfawcett:
Thanks for submitting this. I’m not sure this is the way we should solve it because it just moves the possible crash to another place. The crash is caused because your basal profile is not uploaded by Loop. So under the standard method for usage of the app, this crash shouldn’t ever happen. Loop uploads them as numbers, but if you create the profile in NS, it gets saved as strings. The downside to your change is that it will still crash on invalid values, which can happen more easily due to being typed into NS. It can also cause the graph to crash if there are overlapping values or similar. Loop prevents these from even being saved.

if you’d like to solve this, my suggest is one of 2 things:
- make it completely ignore non-numbers. This solves the crash issue. It will not try to process or graph manual string entries. This is actually how I have done it for manually entered bolus and carbs.
- or still utilize double as is but in a guard or catch that tries your new method if it fails. And if it fails, it will need extra error checking that the string converted value is valid.

PR #182, opened 2023-02-04
Fix Speak BG while adding delta

mjliedtke:
Adjusted where Speak BG code goes to make it trigger when BG is updated.  Also, moved the update badge to the same spot to avoid redundancy.  Other spots in the code actually trigger Speak to go multiple times.


mjliedtke:
Forgot to mention that to get Speak BG working the first time.  You have to iPhone->Settings->Accessibility->Spoken Content->Voices and select or play a voice...not sure why, but I works fine after doing this.

PR #184, opened 2023-04-02
speakbg

bjorkert:
The Speak BG functionality had been experiencing issues for a while, but it has now returned, more powerful than ever. Not only will it announce the present BG value, but it will also provide the difference between the current reading and the previous one.

Verified on iOS 16.4 and macOS

### DIFF
--- a/LoopFollow/Controllers/Alarms.swift
+++ b/LoopFollow/Controllers/Alarms.swift
@@ -848,13 +848,24 @@ extension MainViewController {
     }
     
     
-    func speakBG(sgv: Int) {
-           var speechSynthesizer = AVSpeechSynthesizer()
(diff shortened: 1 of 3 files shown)


mjliedtke:
> The Speak BG functionality had been experiencing issues for a while, but it has now returned, more powerful than ever. Not only will it announce the present BG value, but it will also provide the difference between the current reading and the previous one.
> 
> Verified on iOS 16.4 and macOS

Thanks Jonas.  PR 182 is still sitting out there as well.  I'm pretty novice, so I'll defer to yours.  Doubt mine was working on macOS

bjorkert:
@mjliedtke I think you did a great job, somehow it did not work for me. 

Great to see some interest in LoopFollow development!

mjliedtke:
@bjorkert Cool.  I'll build and test it out this one.

mjliedtke:
@bjorkert Nice tweak with direction.  Love it.  @marionbarker (assuming you are promoting?) tested on 16.4 and macOS as well

PR #186, opened 2023-04-02
Carbs Today

bjorkert:
This pr adds a new item to the 'Information Table' called 'Carbs Today.' This feature provides a sum of all registered carbs since midnight to help you keep track of your child's carb intake for the day. However, please note that this feature may not be useful if you are using fake carbs.


flyingpie101:
I assume this would be an optional switch in the Information Display section ?

bjorkert:
Yes, just as the others. With custom sorting/positioning.

image

PR #187, opened 2023-04-02
Blue Bolus Line -30 minutes and Protein Line -90 minutes

bjorkert:
A blue line is added 30 minutes prior to the current time to provide a clearer view of the boluses that have started to take effect. I described it in this issue, resulted in some modifications but not the blue line. https://github.com/jonfawcett/LoopFollow/issues/110


bjorkert:
Added the 90 minute (protein) line to this PR and added toggles for both. The toggles are default off.

PR #188, opened 2023-04-02
Protein line -90 minutes

bjorkert:
An line is added to the graph 90 minutes prior to the current time to indicate if a meal is causing a blood sugar rise 90 minutes later, which may be due to protein that typically takes about 90 minutes to show its effects.


flyingpie101:
Is there a mockup to see what this would look like ?

bjorkert:
Yes, here is an example of the line:

image

bjorkert:
I have incorporated this feature into #187

PR #194, opened 2023-04-08
pull-to-refresh

### LINKED ISSUE #5: Force refresh option
Add button or drag screen to force a reload of NS data and reset the timer

bjorkert:
This PR introduces a pull-down-to-refresh feature, with the pull-down action implemented on the BG value text. This design choice was made because other controls in the app already involve pulling and scrolling actions. When a refresh is initiated, the feature updates all values by invalidating and resetting all timers, which in turn triggers the download of new data.

Resolves #5


marionbarker:
Slick - very nice. Merging now.

PR #203, opened 2023-07-07
Low Battery Alarm

bjorkert:
This pull request introduces a new low battery alarm feature to LoopFollow. The feature triggers an alarm when the Loop phone battery level falls below a user-specified threshold. This enhancement is aimed at ensuring that users can take timely action to recharge their device's battery, thereby maintaining uninterrupted functioning of Loop.

Changes include:
- A new setting that allows users to specify the battery level at which the alarm should be triggered.
- A mechanism to monitor the device's battery level and trigger an alarm when the level falls below the specified threshold.
- A Snooze functionality for the battery alarm, along with customizable snooze duration.

Looking forward to your feedback on this new feature.

### DIFF
--- a/LoopFollow/Controllers/Alarms.swift
+++ b/LoopFollow/Controllers/Alarms.swift
@@ -496,9 +496,19 @@ extension MainViewController {
             }
         }
         
-        
-        
-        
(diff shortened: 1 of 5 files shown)


marionbarker:
tested and works as expected.

PR #206, opened 2023-07-26
Multiple loopers

bjorkert:
This pull request enables the customization of app names and bundle identifiers for multiple instances of LoopFollow. Currently, the bundle identifier is set using the unique_id variable, which is set to ${DEVELOPMENT_TEAM}. With this change, users can create distinct app names and bundle identifiers.

To use a custom app name and bundle identifier, remove the // comments on the following two lines in Config.xcconfig:
//display_name = Bob
//unique_id = $(display_name)

Replace "Bob" with the desired app name for the instance. For example, setting display_name to "Jonas" will result in a LoopFollow app named Jonas and a bundle identifier of com.Jonas.LoopFollow.

LoopFollow/Info.plist is updated to use the custom display_name in the CFBundleDisplayName. If this value is not set, the app name will default to LoopFollow.

### DIFF
--- a/Config.xcconfig
+++ b/Config.xcconfig
@@ -10,10 +10,10 @@
 unique_id = ${DEVELOPMENT_TEAM}
 
 
-// To build multiple instances of LoopFollow, uncomment the next two lines
-// and replace "Bob" with the desired app name for this instance:
-display_name = Bob
-unique_id = $(display_name)
+// To build multiple instances of LoopFollow, remove the // on
+// the next two lines and replace "Bob" with the desired app name for this instance:
+//display_name = Bob
+//unique_id = $(display_name)
 
 // Explanation:
 // --------------------------------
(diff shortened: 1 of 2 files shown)


bjorkert:
Putting this on hold, I want to synchronize how bundle_id is set with how we do it for the browser build, so an Xcode build can replace a browser build and vice versa.

fzzarf220:
Was the issue ever resolved?  Having ability for multiple loopers would be quite beneficial!!

bjorkert:
> Was the issue ever resolved? Having ability for multiple loopers would be quite beneficial!!

It is still in draft. As it is, it is working fine for Mac-Xcode build but I also want it to cover Browser-Build, thats what delaying it.

fzzarf220:
> > Was the issue ever resolved? Having ability for multiple loopers would be quite beneficial!!
> 
> It is still in draft. As it is, it is working fine for Mac-Xcode build but I also want it to cover Browser-Build, thats what delaying it.

I wait impatiently :)

One thing that would also make this great for browser builds, if we could use Env Variables instead of touching source.  This way no file has to change, just variables set at build time.  Not sure how that could be done, but that would also be powerful!  Just a thought.

PR #207, opened 2023-07-27
Filter out values above 600

bjorkert:
A fix for the issue where user get a 4000 value that messes up the graph and statistics when starting a G7 the first time.


marionbarker:
This does remove values above 3000, but does the Loop prediction get affected? If so, that also messes up graph scaling.

bjorkert:
Limit changed to 600, and now also the prediction is cut if above 600.

PR #208, opened 2023-07-27
Preparation for ios17 calendar access

bjorkert:
This will build with lower versions of xcode since there is a compile time swift version check

Calendar access control is moved centrally to an extension

If the app is missing access, this is now displayed in the apple watch/calendar menu

Calendar access is changed to follow: https://developer.apple.com/documentation/technotes/tn3152-migrating-to-the-latest-calendar-access-levels


marionbarker:
I confirmed it builds, but did not test the calendar features.

marionbarker:
I figured out how to test the calendar. It works (but this is not with iOS 17 or Xcode 15).

aug0211:
Tested on Xcode 15 and iOS 17 and this works. Thanks for the fix!

Heads up, there are some other broken URLs with %20 that need to be updated to + for iOS 17. The latest "Carbs today" call has the working +, but the older calls for SAGE and CAGE still have %20 which will not work.

Figured I'd drop here as a comment in the same "iOS 17 compatibility" spirit.

PR #213, opened 2023-08-15
dexcom update

bjorkert:
Dexcom update.

Repo update to LnL.

Do not merge this one before dexcom-share is merged, after that we should change back to master, run pod install, and commit.


marionbarker:
found 3 more places in README and testflight that needed to be updated from jonfawcett to loopandlearn.
add those changes to this PR.

Issue #134, opened 2021-05-25
Feature Request: Incorporate oref devicestatus into Loop Follow for FreeAPS X Users


sethgagnon:
Requesting that Loop Follow incorporate the ability to read in devicestatus from oref algo for use with FeeAPS X. This will allow for proper COB, IOB, etc display for FreeAPS X users.

jonfawcett:
I'll leave this open, but you'll have to find a FreeAPSX user who is willing to code this.

sethgagnon:
OK. Yea, I can see if I can find someone. I took an export of the devicestatus collection entries (one for Loop and one for FreeAPS X, figuring it would be a mapping exercise, more or less. However, that does not seem to be the case :) I can post those two entries here if that makes it easier for someone to edit existing code you have to accommodate the openaps devicestatus?

scrappy:
I'm confused on this ... I too am using FreeAPS X now, and am just checking whether Follow will work for it ... 

Right now, all of the data that Seth mentions above is uploaded and visible via Nightscout, so if I have Follow pointing to NIghtscout for all of its data, shouldn't it "just work"?   Or does Nightscout just store devicestatus raw in the database and depending on if we have OpenAPS or Loop enabled in settings determines how that data is read in?

To Seth ... can you attach the two devicestatus entries you created so that I can see the difference between the two?

Thx

sethgagnon:
devicestatus from Loop:

[...]

PR #217, opened 2023-09-13
iAPS support for LoopFollow

bjorkert:
- Auto detection of iAPS devicestatus vs Loop
- A new info table category "Autosens" was added.
- infoTable.rowHeight changed from 24 to 21.
- Carbs today is now based on created_at instead of timestamp to fit iAPS
- Ensured that infoVisible values have the correct default settings and length. This fixes a bug that made LF crash for some users at rebuilding.
- The title of a switch for downloading data was updated from "Download Loop/FreeAPS Data" to "Download Loop/iAPS Data".

Many changes are copied/inspired from/by this fork: https://github.com/aug0211/aug0211-LoopFollowFAX/tree/fax

### DIFF
--- a/LoopFollow/Controllers/Graphs.swift
+++ b/LoopFollow/Controllers/Graphs.swift
@@ -548,14 +548,13 @@ extension MainViewController {
         BGChart.moveViewToAnimated(xValue: dateTimeUtils.getNowTimeIntervalUTC() - (BGChart.visibleXRange * 0.7), yValue: 0.0, axis: .right, duration: 1, easingOption: .easeInBack)
     }
     
-    func updatePredictionGraph() {
+    func updatePredictionGraph(color: UIColor? = nil) {
         let dataIndex = 1
         var mainChart = BGChart.lineData!.dataSets[dataIndex] as! LineChartDataSet
         var smallChart = BGChartFull.lineData!.dataSets[dataIndex] as! LineChartDataSet
         mainChart.clear()
         smallChart.clear()
-        if UserDefaultsRepository.debugLog.value { self.writeDebugLog(value: "Graph: print prediction") }
-
+        
         var colors = [NSUIColor]()
         let maxBGOffset: Float = 20
         for i in 0..<predictionData.count {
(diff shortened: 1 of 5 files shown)


bjorkert:
I've identified a regression in the iAPS branch: the Loop overrides are no longer displayed as a banner/stripe.

bjorkert:
Closes https://github.com/loopandlearn/LoopFollow/issues/134

flyingpie101:
Not sure if this is the right place to comment.
1.  The eventual BG display directly under the current BG time since value, is hardcodes to mg/dl and doesnt follow the mmol setting.
image

2. Auggie's version also had a min/max in the information screen on the right.  not as important to me personally, but others might be using it.

bjorkert:
1) Thank you. I'll fix that!
2) Min/Max is now "Pred.", it is re-using the same label as LF used for Loop. I hope that is ok, otherwise it can be improved further on.

flyingpie101:
> 1. Thank you. I'll fix that!
> 2. Min/Max is now "Pred.", it is re-using the same label as LF used for Loop. I hope that is ok, otherwise it can be improved further on.

Pred is fine.  I am sorry I didnt notice it.

bjorkert:
> > 1. Thank you. I'll fix that!
> > 2. Min/Max is now "Pred.", it is re-using the same label as LF used for Loop. I hope that is ok, otherwise it can be improved further on.
> 
> Pred is fine. I am sorry I didnt notice it.

Thank you for the feedback. I've made the necessary adjustments. Could you please build the latest version and check out the fix? Really appreciate your input on this. We're getting close to merging this PR. Anything else you'd like to point out or suggest before we go ahead? Hope everything looks good on your side!

flyingpie101:
Built and testing. So far, so good. 

I do have 1 request that I personally think would be nice.  And I 'bug' that is less important.

Request:
In the OpenAPS pill, there is occasionally a pice of info saying "Additional 13 carbs required with in 10 mins" or whatever the values need to be.   It would be great to be able to see in the informational display of top right of LF a new line saying "Carbs Needed"

Bug:
We did a Dexcom change last night, and after about 30 mins, if I launched LF, it would just crash/close.  Once readings started coming in again, it would launch fine. Not a high priority for me.  But worth mentioning.

bjorkert:
Thanks for the feedback!

I've noted the "Carbs Needed" suggestion and the LF crash issue in my to-do list. Definitely something to look into.

Considering our earlier convo, I think it's best we merge this PR if it aligns with the FAX branch and current LF. We can handle these new points in follow-up updates. Sound good?

Appreciate your input as always!

flyingpie101:
Works for me.

Thx

Greg

PR #219, opened 2023-09-18
Sage>10days

bjornoleh:
LoopFollow does not display SAGE values of more than 10 days. That was fine when everyone was using factory G6 transmitters, which cut the session at ten days sharp. Today, many are using Anubis transmitters which allow (theoretically) running sensors up to 60 days. 

This PR changes the ten day limit into a sixty days limit. 

I have not changed anything related to Sensor change reminders, which still assume ten days is the max, and allows alerts  to be set a user defined number of hours before the ten day limit. In my view, this is fine.

### DIFF
--- a/LoopFollow/Controllers/Nightscout/SAge.swift
+++ b/LoopFollow/Controllers/Nightscout/SAge.swift
@@ -10,7 +10,7 @@ import Foundation
 extension MainViewController {
     // NS Sage Web Call
     func webLoadNSSage() {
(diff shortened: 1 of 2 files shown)


bjornoleh:
Before:
IMG_5680

IMG_5681


After:
IMG_5682

bjornoleh:
Adapted to refactor

marionbarker:
Tested and my SAGE is unchanged. Merge to dev and allow several PR to accumulate in dev before merging to main.

PR #226, opened 2023-10-01
fix for override without duration

bjorkert:
CleanShot 2023-10-01 at 21 08 20@2x

Only the last override is allowed to run without a duration.


dnzxy (approved):
Personally, not a huge fan of index based for-loops, `for…in` is clearer to me and then you could do something like `if current == entries.last`, but that's subjective.

Looks good and quick fix to an issue :)

bjorkert:
> Personally, not a huge fan of index based for-loops, `for…in` is clearer to me and then you could do something like `if current == entries.last`, but that's subjective.
> 
> Looks good and quick fix to an issue :)

I agree. Did not intend do to any refactoring in order to keep things stable and minimize testing for this one.

PR #228, opened 2023-10-05
No fraction digits for COB

bjornoleh:
Do not report fraction digits for COB in the Info Display, as that makes no sense.

Originally suggested by @aug0211, not sure if it was made into a PR before.


bjornoleh:
The PR has been updated to match the refactored code.

marionbarker:
Merge to dev and allow several PR to accumulate in dev before merging to main.

PR #231, opened 2023-10-14
URL fix for iOS 17, replace %20 with + for CAGE and SAGE pulls

aug0211:
SAGE and CAGE URLs must use + instead of %20 for iOS 17


bjorkert:
@aug0211 Ok to close this one since the related code is refactored?

aug0211:
Makes sense if refactor is merged or merging soon 😔 If that’s going to take longer, we should get this fix in faster since anyone on iOS 17 will be impacted.

PR #238, opened 2023-10-14
Refactoring

bjorkert:
I have been using this branch for some time now without issues.

It is mainly refactoring of nightscout downloading, centralizing duplicate code.
Skip redundant download Carbs Today, SAGE and CAGE data. This should save some battery.
Ignores future sensor starts.

There is more to be done in this area but I think it is a good time to merge it and continue from dev. We are getting pull requests as well and it would be good if they are based from code that is not outdated.


flyingpie101:
Installed this branch today.  Looks good so far, except the Carbs Today is stuck at zero

bjorkert:
> Installed this branch today. Looks good so far, except the Carbs Today is stuck at zero

Did you have that issue before, using the dev/main branch? Would you be ok with pm/mailing me the nightscout address (and read-only token if needed) and I can debug using that one?

flyingpie101:
No, I did not have the issue on the Dev branch.  Only once I switch to the refactor branch.
I will PM you my NS on Discord

flyingpie101:
Confirmed. resolved for me

PR #254, opened 2024-01-15
Display profile in infotables

bjorkert:
CleanShot 2024-01-15 at 15 03 58


marionbarker (approved):
Code review looks good.
Tested and it works (shows Default as profile).
Let's wait until the other half (uploading profile names with the Loop customization) is added before merging.

PR #259, opened 2024-03-04
Fix: Prevent crash by handling missing eventType in updateTreatments

bjorkert:
## Summary

This pull request addresses a critical issue where the application could crash if an `eventType` was missing in any of the entries processed by the `updateTreatments` function. By implementing a guard statement to check for the presence of `eventType`, we ensure the function safely skips any entries lacking this essential piece of data.

## Changes

- Added a guard statement in `updateTreatments` to check for the presence of `eventType` in each entry.
- Entries without an `eventType` are now safely ignored, preventing potential crashes.

## Testing

- Tested with a dataset containing a mix of entries with and without `eventType` to ensure that the function behaves as expected.
- Confirmed that entries missing `eventType` are skipped without causing crashes, while valid entries continue to be processed correctly.

Please review the changes and let me know if there are any concerns or further improvements needed. Thank you!

### DIFF
--- a/LoopFollow/Controllers/Nightscout/Treatments.swift
+++ b/LoopFollow/Controllers/Nightscout/Treatments.swift
@@ -50,42 +50,39 @@ extension MainViewController {
         var pumpSiteChange: [cageData] = []
         var cgmSensorStart: [sageData] = []
         
-        for i in 0..<entries.count {
-            let entry = entries[i] as [String : AnyObject]?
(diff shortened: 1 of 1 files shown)


bjorkert:
- Switched from a traditional `for i` loop to a more Swift-idiomatic `for-in` loop.
- Corrected a typo in the condition that checks `basalData.count`.

PR #260, opened 2024-03-05
Add function to clear web cache (Nightscout) when refreshing view by swiping down

dsnallfot:
**Issue**
When applying any kind of code changes in Nightscout, the web cache in Loop Follow tends to keep the old version in cache for a long time. 

There is currently no easy way to clear the web cache to make changes in nightscout visible immediately (other than to delete and reinstall Loop Follow, or wait for an extended period of time (weeks?) until the cache clears automatically)

The everyday user of LF probably dont have any needs to frequently clear the web cache, but for developers and other "super users" the inability to clear the web cache are more troublesome.

**Suggested solution**
This PR introduces an easy way to clear the nightscout web cache by extending the current function reloadWebView to also clear web cache when swiping down.

One potential UX issue with clearing the cache is that all Nightscout settings are reverted to defaults in config vars (ie any changes of settings in the LF Nightscout browser instance gets overridden when clearing the cache). 

To adress this issue an alert _"Do you want to clear the web cache?\nNote: All Nightscout settings will be reverted to defaults!"_ is presented and the user can select yes to continue with clearing the cache, or no to just reload the web view (as current app behavior before this PR).

The PR has been tested in Xcode simulator and on a running iPhone 13, and is behaving as expected.

See attached screen recording for look n feel demo.

https://github.com/loopandlearn/LoopFollow/assets/72826201/245dbc6d-a4e5-4562-8952-c6daab2c3c05

### DIFF
--- a/LoopFollow/ViewControllers/NightScoutViewController.swift
+++ b/LoopFollow/ViewControllers/NightScoutViewController.swift
@@ -64,11 +64,12 @@ class NightscoutViewController: UIViewController {
     // New code to clear web cache
     func clearWebCache() {
         let dataStore = WKWebsiteDataStore.default()
+        let cacheTypes = Set([WKWebsiteDataTypeDiskCache, WKWebsiteDataTypeMemoryCache])
         let date = Date(timeIntervalSince1970: 0)
-        dataStore.removeData(ofTypes: WKWebsiteDataStore.allWebsiteDataTypes(), modifiedSince: date) {
-            print("Web cache cleared.")
+        dataStore.removeData(ofTypes: cacheTypes, modifiedSince: date) {
+          print(“Web cache cleared.“)
         }
-    }
+      }
     
     // this handles target=_blank links by opening them in the same view
     func webView(webView: WKWebView!, createWebViewWithConfiguration configuration: WKWebViewConfiguration!, forNavigationAction navigationAction: WKNavigationAction!, windowFeatures: WKWindowFeatures!) -> WKWebView! {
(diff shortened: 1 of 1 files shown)


marionbarker:
I'm reviewing your request and will go over it with @bjorkert.

We want to maintain LoopFollow to support Loop and iAPS as much as possible.

bjorkert:
Hi, 

What if we only clear the disk and memory cache, preserving cookies? This way, we could avoid the confirmation dialog, making refreshes seamless and safe for everyone.

`func clearWebCache() {
    let dataStore = WKWebsiteDataStore.default()
    let cacheTypes = Set([WKWebsiteDataTypeDiskCache, WKWebsiteDataTypeMemoryCache])
    let date = Date(timeIntervalSince1970: 0)
    dataStore.removeData(ofTypes: cacheTypes, modifiedSince: date) {
      print(“Web cache cleared.“)
    }
  }`
  
  BR/Jonas

dsnallfot:
Thanks Jonas and Marion, 
That absolutely seems like a more elegant solution! 

Do you want to do that change yourself when suitable, or do you want me to update the PR with this code?

marionbarker:
Jonas and I had fun tag-teaming to modify your branch.
I tested this and my custom setting for the NS view in LF remained after doing the clear cache gesture.

@dsnallfot Please test this and if you approve we will go on and merge.

It's now completely silent - pull down in NS view to clear the cache. In Xcode log only, you see Web cache cleared print statement.

dsnallfot:
Great collaboration and quick actio theren! 😊  Will test this right away (but also deploy an NS instance with a few UI test changes to make sure they get updated when clearing cache in LF). Get back to you shortly!

dsnallfot:
Test NS with a few UI changes deployed, your LF code changes tested on Xcode sim build (and my old code tested in parallel on running phone). Same positive result regarding immediately refreshed NS UI. But your more efficient code also maintain all NS settings as expected.

Great job! A+ approved 😊

bjorkert (approved):
LGTM

PR #263, opened 2024-03-16
Improve readability of Snoozer tab, accommodate visual impairment

aug0211:
**Improve readability of Snoozer tab, accommodate visual impairment**

- Make text in the Snoozer view bigger to improve readability
- BG delta, time since last reading, clock time fronts are all >50% larger (40pt --> 65 pt)
- No new data is shown

This is useful in a number of situations, including but not limited to:

1. Visual impairment or those with glasses/contacts who many not always have them on (eg. middle of night)
2. A phone on the edge of the desk, nightstand, table - less squinting and visual strain to see information
3. Phone across the room - eg. unloading groceries, getting ready in the morning, doing some other task - now able to see data from further away

### DIFF
--- a/LoopFollow/Application/Base.lproj/Main.storyboard
+++ b/LoopFollow/Application/Base.lproj/Main.storyboard
@@ -353,7 +353,7 @@
                                         <nil key="textColor"/>
                                         <color key="highlightedColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
                                         <variation key="heightClass=regular-widthClass=compact">
-                                            <fontDescription key="fontDescription" type="system" pointSize="40"/>
+                                            <fontDescription key="fontDescription" type="system" pointSize="65"/>
                                         </variation>
                                     </label>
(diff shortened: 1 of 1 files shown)


bjorkert:
Adding screen shots (before-after)
IMG_3186
IMG_3185

marionbarker:
On a smaller phone (SE):
* The graphic below shows before on the left and after applying this PR on the right

I'm going to accept this for the following reasons:
1. The increased fonts selected are all the same size for the 3 changes (and probably as big as can be)
2. I don't know which of those 3 might be most important for a follower but a user who does use this has selected all 3 as equally important
3. When I take my glasses off and get close enough to resolve the graphic on the right, I cannot resolve the graphic on the left

pr-263-on-se

marionbarker:
While testing this for a longer cycle (stopping the CGM inputs into the test NS site), I discovered a few things I thought should be improved. @bjorkert will be looking into these improvements in the new branch just created.

I changed the PR to point to the new branch and then I will merge it.

Issues to be solved:
* The glucose should not be showing without a strikethrough when stale
* The time ago format should depend on how long it has been, e.g., stop counting seconds at some point
* If you have others, feel free to add a comment for consideration

Issue #258, opened 2024-02-16
Update the build workflow to use macos-13 and Xcode 15


bjornoleh:
We need to update the build workflow to use macos-13 and Xcode 15, as iOS 17 SDK will soon be required at App Store Connect.

A quick overview of SDK versions is available here:
https://xcodereleases.com/?scope=release

GitHub macos-13 runner, Xcode versions:
https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode

Please see email from :

![image](https://github.com/loopandlearn/LoopFollow/assets/63544115/528fcfdf-ba14-4340-82ef-c696d534f296)

Relevant lines:
https://github.com/loopandlearn/LoopFollow/blob/5e3a016b0d80026246286bec25eb26a5b0fe2237/.github/workflows/build_LoopFollow.yml#L23-L27

Probably there are other workflow changes to import from Loop as well, as LF is running a little behind now.

PR #265, opened 2024-03-21
Update to Xcode 15 and iOS 17 SDK for App Store compliance

bjorkert:
Updated the build system to use Xcode 15 and iOS 17 SDK to meet the latest App Store requirements.

### DIFF
--- a/.github/workflows/add_identifiers.yml
+++ b/.github/workflows/add_identifiers.yml
@@ -11,7 +11,7 @@ jobs:
 
   identifiers:
     needs: validate
-    runs-on: macos-12
+    runs-on: macos-13
     steps:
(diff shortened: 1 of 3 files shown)


marionbarker:
I believe this is insufficient. Asking @dnzxy to review. 

See https://github.com/LoopKit/LoopWorkspace/pull/120

dnzxy (approved):
This should be okay. I‘d suggest also bumping all other actions (validate, identifiers, certificates) to macOS 13 runners.  Unfortunately, `macos-14` are still in beta state for GH.

bjorkert:
Thanks @dnzxy, good point! I have pushed that change now!

bjorkert:
Fixes #258

PR #266, opened 2024-03-24
Snoozer improvements

marionbarker:
Start with changes from PR #263. Then add more improvements.
* add strikethrough to font with glucose is stale (>12 minutes, 720 second).


marionbarker:
Initial test are OK for snoozer. But similar strikethrough should be added to Main Screen and Calendar.
Also - the strikethrough should probably follow the Color enable or disable selection found in the General Settings.

lf-pr-266-options

marionbarker:
This is main screen after updating to add strike-out stale glucose values on that screen as well.
Left with `Color Main BG Text` enabled, Right with it disabled.

lf-266-main

PR #267, opened 2024-03-25
Remove Watch Image Feature

bjorkert:
This pull request removes the feature that allowed users to save images of the graph view for the purpose of using them on the watch face. The change includes the deletion of the `UIImageExtension.swift` file and associated code that handled image resizing and saving to the photo library. Additionally, permissions related to photo library access have been removed from the `Info.plist` file.


marionbarker:
* Code review looks good.
* Built on test phone and the rows for saving watch images are gone
   * Verify that Calendar continues to update
* Tested that BG speak still works.

Removed all the old watch photos by deleting them in the Loop Follow and Loop Follow Old albums using the native Photo app on the test phone.

PR #280, opened 2024-04-06
Add UI update for 'Not Looping' status indication

bjorkert:
## Summary
This PR introduces an update to the UI to indicate when Loop/iAPS is 'Not Looping'. The `LoopStatusLabel` is now dynamically updated to expand and fill the entire area designated for loop status messages when no recent loop activity has been confirmed.

## Changes
- Added `evaluateNotLooping` function to determine the 'Not Looping' status.
- UI updates to `LoopStatusLabel` to cover the entire `UIStackView` when not looping.
- `PredictionLabel` is hidden when 'Not Looping' status is active.

## Testing
The changes have been tested to ensure the label updates correctly and the 'Not Looping' status is clear and visible.

### DIFF
--- a/LoopFollow/Controllers/Alarms.swift
+++ b/LoopFollow/Controllers/Alarms.swift
@@ -349,7 +349,7 @@ extension MainViewController {
                 if (UserDefaultsRepository.alertNotLoopingUseLimits.value
                     && (
                         (Float(currentBG) >= UserDefaultsRepository.alertNotLoopingUpperLimit.value
-                            && Float(currentBG) <= UserDefaultsRepository.alertNotLoopingLowerLimit.value) ||
+                            || Float(currentBG) <= UserDefaultsRepository.alertNotLoopingLowerLimit.value) ||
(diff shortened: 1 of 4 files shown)


marionbarker:
Tested Not Looping alarms using test NS and test phone. 

Modified the behavior for `Use BG Limits`. 
* When `Use BG Limits` are enabled:
    * it alarms when glucose is less than value labeled `If Below BG`
    * it alarms when glucose is above value next labeled `If Above BG`

Test Configuration
* I configured glucose to continue uploading to site but stopped the closed-loop upload stream.
* I tested with User BG Limits on and off and with Not Looping alert Active and disabled
* Alarms were as expected (after code modification)

Other comments:
* This phone started in light mode. 
* Then I switched to dark mode.
* The graphic below shows the 2 views. 
 
The Not Looping icon on the main screen is not as obvious in light mode. If we decide to address this, it will be in a different PR.

lf-not-looping-light-dark

PR #281, opened 2024-04-06
Mac catalyst no expiration msg

bjorkert:
## Enhancements

- **Conditional Expiration:** Now hides the expiration section for Mac Catalyst builds, since they are not relevant.


marionbarker:
Tested for Mac Catalyst and iPhone SE. Works as expected.

PR #287, opened 2024-04-27
Replace commas with periods in speech output

bjorkert:
Replace commas with periods in speech output to ensure clarity for users with commas as decimal separators, since commas are interpreted as pauses rather than spoken.


marionbarker:
Code review looks ok and builds fine. My ear cannot tell the difference. Merging into dev.

PR #288, opened 2024-04-28
 Implement Display of Manual BG Values and Standardize Unit Conversion

bjorkert:
This PR introduces the display of manual BG values from Loop and Apple Health in the app's graphs. It also ensures that manual BG check entries from Nightscout using mmol are correctly shown. Additionally, the PR standardizes mmol/L <-> mg/dL conversions throughout the application using centralized constants for improved consistency and maintainability.


marionbarker:
Code review looks good with possible suggestion to later modify the capitalization for some variables: in the US it is mg/dL not mg/Dl. The actual name is not that important, so merging as is.

Confirmed that units display as expected and manual glucose entry (from NS careportal) shows up on LoopFollow display.

PR #290, opened 2024-05-04
Add support for pasting URL with token to populate both fields

bjorkert:
### Summary

This PR enhances the `SettingsViewController` by allowing users to paste a URL that includes a token, automatically populating both the URL and token fields. This improvement aims to streamline the testing process.

### Changes
1. **New functionality:**  
   - Added support for handling URLs with embedded tokens.
   - Automatically extracts and populates the token if found.
   - Cleans up the URL accordingly.
2. **Code refactoring:**  
   - Improved URL filtering and normalization logic.
   - Removed redundant code and improved formatting.

### Testing
1. **Manual Testing:**  
   - Verified that pasting a URL with a token correctly populates the fields.
   - Ensured that normal URL input behavior remains unchanged.
2. **Automated Testing:**  
   - Added relevant unit tests for the new functionality.

### Notes
This update primarily facilitates easier testing scenarios where tokens are used in the URL. It should not affect normal usage but enhances the workflow for testers and developers.

### DIFF
--- a/LoopFollow/ViewControllers/SettingsViewController.swift
+++ b/LoopFollow/ViewControllers/SettingsViewController.swift
@@ -12,8 +12,8 @@ import EventKit
 import EventKitUI
 
 class SettingsViewController: FormViewController {
-
-   var appStateController: AppStateController?
+    var tokenRow: TextRow?
+    var appStateController: AppStateController?
     var statusLabelRow: LabelRow!
 
     func showHideNSDetails() {
@@ -38,7 +38,6 @@ class SettingsViewController: FormViewController {
         
         guard let nightscoutTab = self.tabBarController?.tabBar.items![3] else { return }
         nightscoutTab.isEnabled = isEnabled
-        
     }
    
     // Determine if the build is from TestFlight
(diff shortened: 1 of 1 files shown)


marionbarker:
Tested both in the way @bjorkert intended plus added a password in my Apple Keychain. That worked as well.
Create a new password by typing + in the password tool and give it a name (e.g., LF for NS 1) and then add:
1. website (see notes below)
2. username (This is your URL without the token, e.g., https://yoursite.nightscout.com)
3. password (This is your token, it only needs the alpha numeric string after the hyphen - the descriptive text before that is optional)

Hints about the website:
* Don't put in the https:
* It doesn't have to have anything to do with your Nightscout site
   * if you have another password set where you followed the LoopDocs directions for setting it up, this website has to be something different from what you used for the password that has the API_SECRET; 
   * it does not need to be a valid URL, but can't appear in any of your other passwords.
      * for example `fake.org` is accepted, but can only be used once
      * pick a different fake website if you follow more than one person and want to set up this information in your keychain

PR #292, opened 2024-05-08
Allow bg entries as glucose

bjorkert:
Adding glucose as a allowed key for bg values.


marionbarker (approved):
I tested the NS site with the glucose entries. This PR fixed the issue.
Code review looks good.

PR #297, opened 2024-05-13
Remove Download Loop/iAPS Data toggle

bjorkert:
In an effort to simplify the functionality and improve user experience, we have made a small update to the LoopFollow app. Previously, users had to manually toggle the option to download Nightscout data, dependent on a separate 'Download Loop/iAPS Data' setting. This has been streamlined to enhance efficiency and ease of use.

**Key Change:**
Automatic Data Download: The toggle for 'Download Loop/iAPS Data' has been removed. Now, the app will automatically start downloading Nightscout data whenever a URL is entered. This change ensures that users won't need to manage additional settings to start receiving data, reducing setup complexity and potential user errors.
This update is part of our ongoing efforts to make LoopFollow more intuitive and user-friendly, ensuring that your focus can remain on what's most important—managing health data effectively.

As always, we are eager to hear your feedback as we continue to improve the app. Please share your thoughts and experiences after this update.

Ping @MikePlante1

### DIFF
--- a/LoopFollow/Controllers/Timers.swift
+++ b/LoopFollow/Controllers/Timers.swift
@@ -175,7 +175,7 @@ extension MainViewController {
     @objc func deviceStatusTimerDidEnd(_ timer:Timer) {
         
         // reset timer to 1 minute if settings aren't entered
-        if UserDefaultsRepository.url.value == "" || !UserDefaultsRepository.loopUser.value {
+        if UserDefaultsRepository.url.value == "" {
             startDeviceStatusTimer(time: 60)
             return
         }
@@ -200,7 +200,7 @@ extension MainViewController {
         
         
         // reset timer to 1 minute if settings aren't entered
-        if UserDefaultsRepository.url.value == "" || !UserDefaultsRepository.loopUser.value {
+        if UserDefaultsRepository.url.value == "" {
             startTreatmentsTimer(time: 60)
             return
         }
@@ -225,7 +225,7 @@ extension MainViewController {
     @objc func profileTimerDidEnd(_ timer:Timer) {
         
         // reset timer to 1 minute if settings aren't entered
-        if UserDefaultsRepository.url.value == "" || !UserDefaultsRepository.loopUser.value {
+        if UserDefaultsRepository.url.value == "" {
             startProfileTimer(time: 60)
             return
         }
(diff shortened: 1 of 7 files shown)


flyingpie101:
Sorry.  Why are we removing this toggle ?

bjorkert:
> Sorry. Why are we removing this toggle ?

Hi, thanks for commenting!
It is removed in order to clean up the gui, it is a common error to forget to turn on this setting. I believe it was from a time when LF crashed very easily when there was bad data in Nightscout.
Are you using it, and for what?

marionbarker:
I think we should modify this behavior.
I think NS table should be blank any time the NS URL/token is invalid.
* If NS URL is present and valid, NS table displayed and has data
* If NS URL is present but not valid (e.g. bad token), the NS table is still shown but is empty (glucose coming from Dexcom for this test
* repeat but with no Dexcom credentials, BG shows ---, NS table displayed but blank
* NS URL is blank, then NS table is not displayed

flyingpie101:
I have it toggled on, because I use NS and want the NS data downloaded.   
So, I assume this is proposing to remove the toggle and the download data (if available) is always 'enabled' ?
Fine with me if thats the case.

marionbarker:
And while you are at it. This is an unexpected consequence of a PR a couple of times ago where you sanitized the URL.

Try typing in https://sitename, instead of pasting in a URL, and you'll notice it won't allow you to enter the `/` characters.

bjorkert:
> I have it toggled on, because I use NS and want the NS data downloaded. So, I assume this is proposing to remove the toggle and the download data (if available) is always 'enabled' ? Fine with me if thats the case.

Yes, exactly. My fault, I should have written a better PR description of this. So after this change, it will always try to download Nightscout data, just as if the switch would have been on.

flyingpie101:
> And while you are at it. This is an unexpected consequence of a PR a couple of times ago where you sanitized the URL.
> 
> Try typing in https://sitename, instead of pasting in a URL, and you'll notice it won't allow you to enter the `/` characters.

I feel like, perhaps, your comments are for a different PR ;)

bjorkert:
> > > I have it toggled on, because I use NS and want the NS data downloaded. So, I assume this is proposing to remove the toggle and the download data (if available) is always 'enabled' ? Fine with me if thats the case.
> > 
> > 
> > Yes, exactly. My fault, I should have written a better PR description of this. So after this change, it will always try to download Nightscout data, just as if the switch would have been on.
> 
> Sure, I figured, just wanted to make sure. I also assume that if there is no NS url entered. it won't attempt to download every few minutes ?

Yes, blank URL would be the same as toggling the removed toggle off.

marionbarker (approved):
In addition to the successful test that the slider used to enable NS download is no longer needed, this PR also fixes a bug introduced in PR #290. It is now possible to type the `https://` into the URL row.

PR #304, opened 2024-06-15
Trio predictions

bjorkert:
Adds all 4 predictions from Trio
Code refactoring
Fix for transparent tabbar in settings view. Thanks to @petwall for this suggestion.


marionbarker:
The transition between a variety of Trio sites works perfectly in updating the prediction plot.
The transition from Trio to Loop updates all except the prediction. 
* I can force it to update the prediction by dragging down on the Glucose value
* Wait for the next CGM update, the prediction line updates at that point

marionbarker (approved):
tested with latest changes. Works as expected.

PR #305, opened 2024-06-19
Differentiating the display of SMBs from Manual boluses.

bjorkert:
SMB's are now displayed with a light blue dot.
CleanShot 2024-06-19 at 21 40 21@2x
CleanShot 2024-06-19 at 21 42 58@2x


marionbarker:
Tested with Trio (left) and Loop (right) examples. 
This looks good to me. Merging to dev.

In both cases:
* there are carbs in the past and in the future
* there are both manual and automatic boluses
* the indicators no longer fade over time

lf-future-carbs-trio-loop

marionbarker (approved):

PR #306, opened 2024-06-22
Alert if the app is not active in the background

bjorkert:
Shows an alert if the app hasn't been active for 6 minutes in background mode. BG is checked at least every 5 minutes.


marionbarker:
## Summary
Success, merging to dev

## Testing
Build with this branch
* make sure data is coming in from data source
* return to home screen
* quit the app, lock the phone
* wait at least 10 minutes in each case
* with Notifications enabled (and persistant) for LoopFollow (in iOS phone settings)
   * Notification is visible warning that LoopFollow is not running (graphic below)
* with Notifications disabled for LoopFollow (in iOS phone settings)
   * nothing happens
* enable Notifications; must open LoopFollow one time, then can repeat the test and get the alert (phone locked or unlocked)

loopfollow-background-alert

Issue #121, opened 2021-01-19
Graph is not rescaled after high predicted values disappear


yuri-zhuk:
*[Environment]*
iPhone 12 Pro, iOS 14.3

*[Preconditions]*
Glucose entry (e.g. lunch) is added 4 hours into the future without bolus - predicted BG very high (~30 mmol).

*[Steps]*
Wait for the lunch, giving bolus as appropriate.

*[Actual result]*
The graph area in Loop Follow remains scaled for very high values, even though they are no longer present anywhere on the graph.

*[Expected result]*
The graph area is scaled back to whatever is the present highest value on the graph.

*[Workaround]*
Restarting the app fixes it.

marionbarker:
## Status - this still needs to be fixed

Using latest LoopFollow (dev branch, commit 57e6350)

* I entered a bogus value of 399 into Apple Health: Loop, NS and LoopFollow picked it up - graph autoscaled
* I removed the bogus glucose value: Loop, NS and LoopFollow all show plots without the bogus value but the autoscaling still accommodates the bogus value
* waited for a few cycles and pulled down glucose - neither action changed the scaling
* restarted the app and scaling updated

PR #307, opened 2024-06-24
Reset topbg for each cycle

bjorkert:
An attempt to solve https://github.com/loopandlearn/LoopFollow/issues/121


marionbarker:
Refer to https://github.com/loopandlearn/LoopFollow/issues/121#issuecomment-2178945353
This indicated that user had to restart the app to get the scaling to restore.

Test by add and then remove fake glucose (of 399):
* LoopFollow showed the fake glucose and autoscaled
* Removed the fake glucose and the scaling updated without needing to touch the app

lf-pr307-reset-scale

marionbarker (approved):
Code review looks good. Test was successful.

Issue #303, opened 2024-06-12
Loop Follow crashes after successful build


reneeloop:
Github.com/Reneeloop/LoopFollow is my GitHub address 

My LoopFollow was not displaying my current Dexcom readings, saying it was so many minutes delayed

I assumed I needed to rebuild via browser. I had a successful browser rebuild of LoopFollow last night and today, but every time I open TestFlight, it opens LoopFollow, but doesn’t display a reading and crashes. 
Yes I signed Apple agreement

PR #308, opened 2024-06-24
Hardened dexcom share and removed dead code

bjorkert:
let data = result!
and
let latestDate = data[0].date

Would crash if data is nil when error is nil, I did refactor the code to avoid a potential crash here in response to: https://github.com/loopandlearn/LoopFollow/issues/303

Also found some dead code that is now removed.


marionbarker (approved):
Did code review - looks nominal.
Built and tested with Loop and Trio users, with and without Dexcom share.

PR #309, opened 2024-06-25
Show the eventually value and loopstatus

bjorkert:
CleanShot 2024-06-25 at 20 47 07


bjorkert:
CleanShot 2024-07-01 at 19 43 09@2x

marionbarker:
Tested and works as expected for both Loop and Trio. 
* The label for the table for prediction min/max is now explicitly labeled min/max
* The value under the current glucose is the eventually value for glucose

updated-display

marionbarker (approved):

PR #314, opened 2024-07-06
update runners to macOS 14 in yml files

marionbarker:
Update all runners to use macOS 14
Update the explicit Xcode version to 15.4


bjorkert (approved):

dnzxy (changes requested):
Let’s please make the same xcode-select changes here and comment out for the build yaml.

marionbarker:
Tested with the explicit choice of Xcode 15.4 as per method used by LoopWorkspace.
Successful test of all actions.

PR #315, opened 2024-07-10
Correct DeviceStatus Download Count and Dynamic BGData Download Limit

bjorkert:
**1.	DeviceStatus.swift:**
	•	Corrected the count parameter for device status downloads from 288 to 1, as only one record is needed. The initial value was a copy-paste error, which led to unnecessary data being downloaded.
**2.	BGData.swift:**
	•	Adjusted the count parameter to dynamically calculate the limit of BG values to download. The new formula accounts for both Loop/Trio and Dexcom Share data, leading to 2 * 288 BG values per day.
	•	Previous hardcoded value: 1000. Two days of double data results in 1152 values.
	•	New dynamic value: UserDefaultsRepository.downloadDays.value * 2 * 24 * 60 / 5


marionbarker:
The code review looks good.
Tested and it works.
Used Trio and Loop NS sources and increased display days to 4. All data is displayed.

Future question:
Q: Do we need to future proof this for a 1-minute CGM update rate? 
A: Not at this time and it would be a separate PR.

marionbarker (approved):

PR #318, opened 2024-07-18
Info Table improvements

bjorkert:
New Info Table Items:
	•	Target: Displays the current target. If there is an override or temporary target affecting the target, it will show “profile target -> effective target.”
	•	ISF (Insulin Sensitivity Factor): Displays the current ISF. If there is an override, temporary target, or autosens affecting the ISF, it will show “profile ISF -> effective ISF.”
	•	CR (Carb Ratio): Displays the current carb ratio. Shows “profile value -> effective value” if they differ.
	•	Updated: Indicates when the data was uploaded to Nightscout.
	•	TDD (Total Daily Dosage): Shows the total daily dosage (currently not uploaded to Trio dev).

Modified Info Table Items:
	•	Basal: If there is a temporary basal, it will show “profile basal -> current basal.” This applies to both Loop and Trio users.

Behind the Scenes:

This pull request includes significant refactoring:

	•	Introduced HKQuantity for blood glucose values, applied to most of the info table code.
	•	Replaced magic numbers for updating info table rows with an enum.
	•	Centralized the updating of the info table to a manager.
	•	Removed unnecessary UserDefaults values.
	•	Replaced instances where insulin and carbs was stored as strings with a Metric class.
	•	Introduced a ProfileManager to keep track of profile values.


marionbarker (approved):

marionbarker:
Arg - forgot that I asked you to rebase.

PR #323, opened 2024-08-06
Add IAGE Support and Migrate InfoDisplaySettings to SwiftUI

bjorkert:
## Summary

This pull request adds support for the IAGE event type and enhances the Information Display Settings with SwiftUI.

### Key Features

1. **IAGE Event Integration**
   - Implemented support for tracking and displaying the IAGE (Insulin Cartridge Change) event.

2. **Information Display Settings Enhancements**
   - Migrated the settings page from UIKit/Eureka to SwiftUI
   - Fixes a bug in the current dev branch where toggling an item's visibility would visually reset the sort order, even though the persisted sort order was correct.


marionbarker:
## Summary
A few minor comments.

## Code Review
Looks good

## Test
Built onto SE
Reading data from OpenAPS (Trio) Test Site
* add treatment for insulin cartridge change and sensor start
* both are displayed on LF
* moved the rows around and info table rearranged as expected

Switch to Loop Site
* Updated is not showing a value - should that not be last Loop upload?

## Suggestions
* Update in table shows 
   * hh:mm:ss am/pm for 12 hour time phone (pretty crowded)
   * hh:mm:ss for 24 hour time display
   * I think hh:mm should be sufficient

marionbarker (approved):
I noted a few changes you may want to address.
Otherwise, I approve the code.

bjorkert:
I can address Updated in a separate PR, since it was introduced in this one.

PR #326, opened 2024-09-24
Filter out cal entries

bjorkert:
A small fix to filter out cal type entries from the bg result from Nightscout.

These entries would stop the bg value to be visible until they are out of scope for the fetch of data from Nightscout.

Preferably we should in the code filter out any type we don't want to use but this is what I have bandwidth to do right now.


marionbarker (approved):
Tested with a URL that had "cal" type (for Libre).
Did not update with main.
Nominal behavior with filter_out_cal

PR #335, opened 2024-12-06
TRC debug info

bjorkert:
Added information about device token, production env, team id and bundle id in the bottom of the Remote/TRC screen.

I verified that switching build method for Trio changes the production env boolean without restarting LoopFollow. Please note, the production env bool is in the profile and the profile information is refreshed every 10 minutes.

CleanShot 2024-12-06 at 21 12 42@2x

### DIFF
--- a/LoopFollow/Remote/Settings/RemoteSettingsView.swift
+++ b/LoopFollow/Remote/Settings/RemoteSettingsView.swift
@@ -192,6 +192,13 @@ struct RemoteSettingsView: View {
                         Toggle("Meal with Fat/Protein", isOn: $viewModel.mealWithFatProtein)
                             .toggleStyle(SwitchToggleStyle())
                     }
+
+                    Section(header: Text("Debug / Info")) {
+                        Text("Device Token: \(Storage.shared.deviceToken.value)")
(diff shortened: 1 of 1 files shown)


marionbarker:
## Status

Initial test is good. Hold on merging until I get a version of Trio-dev that uses the production environment.

## Testing

First I made sure the LoopFollow `dev`, commit 0d79ca9 was working with Trio-dev commit `5afb7e4b1`, created from Xcode.
Then built this version
The Debug display is present and Production Env is false.

Make sure remote commands go through as expected.

Install the most recent Trio-dev from TestFlight (older version) `fbdc3a2`
It does not update the Production Env to true and the remote commands do not go through but then it is an older version of TestFlight.

Ask for help on getting a production version ready to test.

marionbarker (approved):
Approve the changes but want to test with a Production Environment before merging.

marionbarker:
I read the fine print in the original comment.
* I loaded from TestFlight again
* I changed a basal rate, CR and ISF to make sure the profile updated for NS
* I waited long enough (>10 minutes) for the LoopFollow to pick up the information from the new profile
* Production Env now reads true
* Remote commands went through promptly.

marionbarker (approved):

PR #336, opened 2024-12-06
Align Carb and Bolus Guardrail Behavior

bjorkert:
Before, the last valid value was used for the meal entry. Now the invalid value is cleared and the error message is displayed correctly.


marionbarker:
## Status
LGTM

## Test

The setting for max carbs (protein/fat) was set to 30 g.

### Test before

Confirmed that a carb entry of 45 was automatically modified to 4 g and uploaded.

### Test with PR

Confirmed that a carb entry of 45 caused an out-of-bound alert whether hitting the end entry or the send meal button.

Confirmed that when Protein/Fat was added to meal options, they also alerted properly.

marionbarker (approved):

PR #337, opened 2024-12-06
Require auth for meal bolus

bjorkert:
With this change, a authentication is required for bolusing with meals. When there is no bolus, no auth is required.

### DIFF
--- a/LoopFollow/Remote/TRC/MealView.swift
+++ b/LoopFollow/Remote/TRC/MealView.swift
@@ -8,6 +8,7 @@
 
 import SwiftUI
 import HealthKit
+import LocalAuthentication
 
 struct MealView: View {
     @Environment(\.presentationMode) private var presentationMode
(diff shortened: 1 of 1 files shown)


marionbarker:
## Status
LGTM

## Test

Enter carbs, protein, fat and bolus for a meal.

### Test before

Confirmed that no authorization was required for the bolus.

### Test with PR

Confirmed that authorization was required when meal includes a bolus.
* limits are checked first
* a confirmation screen is shown for the desired meal
* the user must enter authorization
* carbs and bolus arrive at NS (with fat/protein spread out of future carbs) and on Trio phone

Confirmed that authorization was **not** required when meal did not include a bolus.
* limits are checked first
* a confirmation screen is shown for the desired meal
* when no bolus is requested, remote command is issued after confirmation

marionbarker (approved):

PR #338, opened 2024-12-06
Clarification about scheduled meal and bolus

bjorkert:
Adding a clarification about sheduling meal and bolus.

CleanShot 2024-12-06 at 22 01 37@2x


marionbarker:
## Status
LGTM

## Test

send a meal with a bolus for a time in the future

### Test before

Confirmed the command is sent but there is no warning that only the carbs are entered in the future; that the bolus is immediate.

### Test with PR

Confirmed that as soon as the Schedule for later is enabled, the user is warned as shown in the graphic above.

marionbarker (approved):

PR #339, opened 2024-12-07
Meal seconds from current time

bjorkert:
Use seconds from current time when scheduling meals. This is a trick to avoid Nightscout to overwrite meals scheduled at the same time.

### DIFF
--- a/LoopFollow/Remote/TRC/MealView.swift
+++ b/LoopFollow/Remote/TRC/MealView.swift
@@ -237,9 +237,10 @@ struct MealView: View {
             let calendar = Calendar.current
             let now = Date()
             let selectedDateComponents = calendar.dateComponents([.hour, .minute], from: selectedTime)
+            let currentSecond = calendar.component(.second, from: now)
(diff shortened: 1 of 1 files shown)


marionbarker:
## Status
LGTM

## Test

Issue two meals in the future at the same time in minutes and hours. Choose a time just a few minutes in the future so can observe the entries in Trio.

### Test before

Confirmed when two carb entries are issued remotely using the same hh:mm future time:
* Nightscout only keeps the second value and discards the first entry
* LoopFollow reads from NS so has the same pattern
* Trio records both correctly (after the hh:mm is past)

### Test with PR

Confirmed when two carb entries are issued remotely using the same hh:mm future time:
* Nightscout records both correctly (there is actually a 1 in 60 chance of hitting exactly the same second, but this is good enough)
* LoopFollow has no provision to show 2 carb entries at the same time as a sum, so it only shows the most recent one on the plot, but the COB is correct.
* Trio records both correctly (after the hh:mm is past)

marionbarker (approved):

PR #340, opened 2024-12-07
Align temp target guardrails with Trio

bjorkert:
This PR aligns the temp target guardrails with trio 80-200 (4,5-11,1).

### DIFF
--- a/LoopFollow/Helpers/TextFieldWithToolBar.swift
+++ b/LoopFollow/Helpers/TextFieldWithToolBar.swift
@@ -177,13 +177,27 @@ public struct TextFieldWithToolBar: UIViewRepresentable {
                         self.parent.quantity = quantity
                         textField.text = self.format(quantity: self.parent.quantity, for: self.unit)
                     } else {
+                        let formatter = NumberFormatter()
+                        formatter.minimumFractionDigits = self.unit.preferredFractionDigits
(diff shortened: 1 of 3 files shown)


marionbarker:
## Status
LGTM

## Test

The oref algorithm has hard limits of 80 to 200 for TempTarget ranges. Previously these could be set from 60 to 300 in LoopFollow. Modify the guardrails to 80 to 200 for LoopFollow. Test both mg/dL and mmol/L settings.

### Test before

* Confirmed that a TT with a target of 60 could be issued.
   * shows up in NS
   * see the 60 on the Trio plot, but the banner says no active adjustment 
   * cancel it from LF, shortened bar on NS / LF / Trio

* Confirmed that a TT with a target of 300 could be issued.
   * shows up in NS
   * see the 300 on the Trio plot, but the banner says no active adjustment 
   * cancel it from LF, shortened bar on NS / LF / Trio

* Repeat with TT of 180 mg/dL

### Test with PR

* Attempt to exceed guardrails and get alert message at 300 mg/dL and 60 mg/dL. 200 and 80 are successful. 79 and 201 show an alert message and will not be accepted.
* Switch units to mmol/L. Low level is now 4.5 mmol/L. High level is 11.1 mmol/L.

marionbarker (approved):

PR #341, opened 2024-12-09
Fix crash for long lines without spaces

bjorkert:
Fix a crash that occurred when wrapping very long words by ensuring the substring length is never negative. This change adjusts the calculation of availableSpace and adds checks to prevent out-of-bound string indexing, stabilizing text wrapping behavior.


marionbarker (approved):
Tested using a site which had a "problem" note. This PR fixes it.
Code review looks good.

PR #342, opened 2024-12-10
Contact image with bg information

bjorkert:
LoopFollow contact image with bg info, intended for apple watch - just add the LoopFollow - BG contact to the watch face to follow the bg value in real time.
* Options for delta value or arrow
* Dynamic font size
* Contact name is based upon the app display name, making it possible to use three instances of LoopFollow showing three different bg values on the watch face.

image

B75575D3-AB6D-48EB-A671-9C07F5050DA8_4_5005_c


marionbarker:
Tested and works nicely.

marionbarker (approved):
Code review looks good.
Tested for mmol/L as well as mg/dL.

PR #344, opened 2024-12-18
Trio remote controll feature toggle

bjorkert:
Adds a temporary modification making TRC only available when not building the Main branch.


marionbarker (approved):
Approved from code review and test.

PR #345, opened 2024-12-22
Fallback on suggested if enacted is not present

bjorkert:
With this change, the suggested record will be used instead of Fallback on suggested if enacted is not present if Fallback on suggested if enacted is not present is not present.

See discussion in discord/tcd-general.


dnzxy (approved):

bjorkert:
@marionbarker I have been using this for a few days and it has been working well for me. I think we can merge.

marionbarker:
This is being used by people who are caregivers. Test that it builds successfully.

Quick test. 
* test phone is running TCD with dana pump using dana simulator
* set Not Looping alarm for 10 minutes
* interrupt the simulator and get the audio alarm

PR #347, opened 2024-12-28
Fix crash by safely handling optional casting for ‘duration’

bjorkert:
This PR addresses a crash caused by force unwrapping during the casting of ‘duration’. The casting is now handled safely using optional binding to prevent runtime errors.


marionbarker (approved):
Approved by code inspection.
Built on test phone. App works normally.

PR #350, opened 2025-01-01
Fix date parsing for Temp Basals using ISO8601DateFormatter

bjorkert:
Updated date parsing for basals using a new helper function supporting both Trio and Loop date formats.

### DIFF
--- a/LoopFollow/Controllers/Nightscout/Treatments/Basals.swift
+++ b/LoopFollow/Controllers/Nightscout/Treatments/Basals.swift
@@ -10,31 +10,6 @@ import Foundation
 
 extension MainViewController {
 
-    private func parseDate(_ rawString: String) -> Date? {
-        var mutableDate = rawString
-
-        if mutableDate.hasSuffix("Z") {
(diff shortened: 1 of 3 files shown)


marionbarker (approved):
1. Code review
   * looks good
   * reviewed again after second commit

1. Test
   * tested with a site that would not display basals; with this fix, the basals are displayed properly
   * first commit failed with PR for a Loop URL; after adding second commit succeeds with Loop URL
   * success for Trio URL for 0.2.2 and 0.3.0

marionbarker:
Test again after additional update that cleans up code even more. This was triggered by noticing a print statement from `LoopFollow/Controllers/Nightscout/Treatments/Carbs.swift`. @bjorkert then moved "the code from the new function to the global one (NightscoutUtils.parseDate)". This is a better implementation that improves date parsing for carbs as well as basal. (Other treatments already use this function).

1. Code review (over all commits): approved
2. Tests successful for
   * URL that had initial problem displaying basals
   * Loop URL
   * Trio 0.2.2 and 0.3.0 URLs

PR #362, opened 2025-01-27
Enable Setting and Canceling Loop Overrides

bjorkert:
This pull request introduces support for setting and canceling Loop overrides.

Requirements:
	•	Use Remote Type Nightscout
	•	The token must have admin authorization (caregiver access is not sufficient).
	•	APNS must be properly configured in Nightscout.
	•	Setting overrides from Nightscout must function as expected.
	
image
image


marionbarker:
Preliminary test. Requested a few feature updates - Jonas responded with a list:

TODO:
[ ] If TRC and not detecting Trio, set remote type to None.
[ ] More user friendly error message for unauthorized, might also detect that the user does not have Admin to start with.
[ ] More user friendly error message for badDeviceToken, hint about production vs sandbox depending on build method.

marionbarker:
## Status

Things that need to be fixed:
* The mini hint should have words "and Loop" added to read: "Nightscout is the only option for the released version of Trio **and Loop**".
* When URL is fed from Loop (and if possible Trio 0.2.x), the Remote Type option for TRC should be disabled
* When URL is fed from Trio-private-beta, changing the Remote Type option from Trio Remote Control to None, the TRC 4-grid of choices was still offered when tapping remote on the tool bar

Optional Things:
* When URL is fed from Loop:
   * might want to include `LOOP_PUSH_SERVER_ENVIRONMENT` in the error message when NS was not configured properly (I'll put something in the code review with a specific suggestion)
   * perhaps a note could be added between the Active Override and Available Override sections that says - "Setting a new override while one is active will cancel the active override and set the new one." Otherwise, we'll be answering a lot of questions on FB
   * Perhaps add word Trio or Loop into token insufficient error messages

## Test

Build branch loop-override-pr, commit a9757f3 on several test phones.

### Test Loop Remote

Use test phone that is viewing a Loop URL, set token to admin
Modify choice for remote type
* Select None
   * Tap on Remote and get "Please select a Remote Type in Settings" 
* Select Trio Remote Control - I don't think that option should be allowed
   * **User is shown screen with places to fill out all the items needed for TRC**
   * Tap on Remote and get "Trio Remote Control is only supported for Trio" 
* Select Nightscout
   * Attempt to set an override (can see the overrides, but not set one)
   * **Better error message than before but maybe it should include `LOOP_PUSH_SERVER_ENVIRONMENT`**
   * Replace the Browser Build Loop with Xcode Loop on the Loop phone
   * Attempt to set an override - got a success message and it shows up on Loop phone
   * Attempt to cancel the override - that also worked
   * Set override on the Loop phone, wait for it to show up on LF phone
   * Set a new override on top of the acting one - it accepted that - action was to remove the old one and set the new one
   * Acceptable behavior but perhaps a note could be added between the Active Override and Available Override sections that says - Setting a new override while one is active, will cancel the active override and set the new one.
* Switch to another URL where the token is readable
   * Tap on remote in tool bar and be told explicitly that token must include admin

### Test Trio Remote using 0.2.3

Use test phone that is viewing URL fed by Trio 0.2.3 app, set token to admin
Start with Trio phone set with Fetch and Remote commands disabled.
Configure LoopFollow phone as below
* Set Remote settings to None
   * Tap on Remote and get "Please select a Remote Type in Settings" 
* Select Trio Remote Control - I don't think that option should be allowed
   * **User is shown screen with places to fill out all the items needed for TRC**
   * Tap on Remote and get "Trio Remote Control is only supported for Trio" 
* Select Nightscout
   * Attempt to cancel a Temp Target currently running - accepted by Nightscout, but Trio phone is not configured for remotes
   * Both LoopFollow and Nightscout think TT was canceled but Trio phone is still running it.
   * Now enable both Fetch Treatments and Remote Control on the Trio phone
   * The TT is now canceled
   * Send named TT from LoopFollow phone, seen on Nightscout and shortly thereafter on the Trio phone
   * Cancel on Trio phone and it updates in Nightscout and then in LoopFollow
* Switch to another URL where the token is readable
   * Tap on remote in tool bar and be told explicitly that token must include both careportal and readable 

### Test Trio Remote using private beta

Use test phone that is viewing a Trio-private-beta URL and is configured to work with the Trio Remote Commands option - make sure it still works with this branch
* Test send a bolus - sent without errors to Nightscout, but Trio phone has Enable Remote Control disabled (see note in NS treatments and if tap dot on LF screen)
* Set Remote settings to None
   * **It should no longer show me options for using TRC but they are shown**
* Set Remote settings to Nightscout
   * Now I am shown the expected named presets and option for custom TT
   * Enter custom TT and it appears on Trio phone
   * These work regardless of Trio feature setting
* Set Remote settings to Trio Remote Commands
   * Enable Remote Control on the Trio phone and make sure shared secret matches
   * Test send a bolus and see bolus enacted on the Trio phone

LoopFollow/Remote/Loop/LoopNightscoutRemoteView.swift
```diff
@@ -0,0 +1,38 @@
...
+
+struct LoopNightscoutRemoteView: View {
+    @Environment(\.presentationMode) var presentationMode
+    @ObservedObject var nsAdmin = ObservableUserDefaults.shared.nsWriteAuth
+
+    var body: some View {
+        NavigationView {
+            if !nsAdmin.value {
+                ErrorMessageView(
+                    message: "Please update your token to include the 'admin' role in order to do remote commands."
```
marionbarker, line 19: add word Loop because this is specific to Loop.
find the equivalent for Trio and add the word Trio (careportal and readable is sufficient)

LoopFollow/Remote/Loop/LoopOverrideView.swift
```diff
@@ -0,0 +1,222 @@
...
+    var body: some View {
+        NavigationView {
+            VStack {
+                if device.value != "Loop" {
+                    ErrorMessageView(
+                        message: "Remote commands are currently only available for Loop."
+                    )
+                } else if !nsAdmin.value {
+                    ErrorMessageView(
+                        message: "Please update your token to include the 'admin' role in order to do remote commands."
```
marionbarker, line 52: add word Loop because this is specific to Loop.
find the equivalent for Trio and add the word Trio (careportal and readable is sufficient)

marionbarker (changes requested):
Please see previous comments for changes I think should be included.

LoopFollow/Remote/Loop/LoopOverrideViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+    }
+
+    func formatErrorMessage(_ errorTitle: String) -> String {
+        switch errorTitle {
+        case "Unauthorized":
+            return "Unauthorized, verify that your token is correct and has admin auth"
+        case "APNs delivery failed: BadDeviceToken":
+            return "APNs delivery failed: BadDeviceToken, verify that the production setting or Browser/XCode build setting is correct in your Nightscout setup."
```
marionbarker, line 77: Proposed rewording:

APNs delivery failed: BadDeviceToken, verify that the LOOP_PUSH_SERVER_ENVIRONMENT parameter in your Nightscout setup matches the build method used for Loop.

bjorkert:
I have addressed all issues except disabling the TDC option for Trio 0.2.x. AFAIK there is no way to distinguish between the 0.2 and 1.0 version using data from Nightscout.

marionbarker (approved):
All requested changes made, reviewed and tested.

Issue #354, opened 2025-01-19
Adapt display of Trio infinite overrides (30 days uploaded to NS)


bjornoleh:
LF display of indefinite overrides is a little broken now. TCD recently (?) changed the duration that is stored and uploaded, which is 30 days. 

![Image](https://github.com/user-attachments/assets/6c64a31f-0e46-474f-9b84-06f166be7101)

marionbarker:
Issue solved with PR #363.

PR #363, opened 2025-01-28
Limit displayed override duration to 'Hours of Prediction' after current time

bjorkert:
Solves https://github.com/loopandlearn/LoopFollow/issues/354

This PR limits displayed override duration to 'Hours of Prediction' after current time.


marionbarker:
## Status

This PR fixes the problem reported in Issue #354

## Code Review

The proposed change will use duration of prediction for showing the currently acting override. This will limit how far in the future all overrides will be shown. Especially important for Trio (soon to be released as 1.0) which uses 30 days to indicate an indefinite override duration.

## Test

Begin without the modification in this PR.
Set an infinite duration on the Trio phone that feeds the URL.
Build onto phone with the modification in this PR.

Graphic below shows before applying the PR on the left and with the PR on the right.

lf-pr-363

marionbarker (approved):
LGTM

PR #364, opened 2025-01-28
Prevent Rec. Bolus Alarm from Repeating

bjorkert:
Solves: https://github.com/loopandlearn/LoopFollow/issues/361

This PR ensures the Rec. Bolus alarm triggers only once per bolus value instead of repeating each time alarms are evaluated.

Problem Before:
	•	The alarm triggered every time alarms were evaluated as long as currentRecBolus >= alertAtRecBolus.
	•	Even if the bolus did not change, the alarm kept playing.
	•	This resulted in unnecessary repeated alarms unless manually snoozed.

Fix:
	•	Introduced lastRecBolusTriggered to track the last alarmed bolus value.
	•	The alarm only triggers if currentRecBolus has changed.
	•	If currentRecBolus drops below alertAtRecBolus, it resets so that future increases will trigger correctly.

### DIFF
--- a/LoopFollow/Controllers/Alarms.swift
+++ b/LoopFollow/Controllers/Alarms.swift
@@ -514,7 +514,7 @@ extension MainViewController {
             let alertAtRecBolus = UserDefaultsRepository.alertRecBolusLevel.value
 
             if currentRecBolus >= alertAtRecBolus {
-                if Observable.shared.lastRecBolusTriggered.value != currentRecBolus {
+                if currentRecBolus > (Observable.shared.lastRecBolusTriggered.value ?? 0) {
                     AlarmSound.whichAlarm = "Rec. Bolus"
 
                     if UserDefaultsRepository.alertRecBolusRepeat.value {


marionbarker:
## Code Review

The majority of the diffs are from reformatting the code.
The substantive change is the addition of Observable.shared.lastRecBolusTriggered.value, updating that value when it changes and alarming only once when it is new.

## Test

Test LoopFollow with this modification
* Set looping phone to disable closed loop so no automatic boluses happen.
* Use flat glucose trace at desired target
* Enter carbs with no bolus to generate a rec. bolus amount (5.6 U)
* Configure LoopFollow phone for Rec. Bolus alert when Rec. Bolus > 1.0 U
* Wait for next loop: get the alarm, plays continuously until hitting Snooze
* Wait for next loop: loop phone shows rec bolus of 5.55 U now - it is different so get the alert again
* And another alert when rec bolus is 5.5 U.

For this situation, the alert still happens every loop. I don't think this is the desired behavior.
* I think if Observable.shared.lastRecBolusTriggered.value is not nil, then the != should be >= when deciding to turn on the alarm.
* that way if the currentRecBolus exceeds the lastRecBolusTriggered.value you get a new alert, but if it is less than lastRecBolusTriggered.value, no new alert is trigger

I'll leave it up to you to modify the logic with proper swift coding if you think it necessary.

bjorkert:
Thanks for the review! I’ve updated the condition to only re-trigger the alarm if the recommended bolus increases beyond the last triggered value.

marionbarker:
## Status

Probably good enough to merge. There may be other changes to be added, but it will solve #361.

## Code review

Last commit looks perfect to ignore smaller rec bolus amounts while allowing an increased amount of rec bolus to alert.
* The sound did not alert when phone was locked with this PR but did with dev (see Test 3), but when I repeated the test with dev it took ~14 minutes for the sound to start the second time - might not have waited long enough when testing locked phone with this PR
* With dev the alert notification on locked phone happened repeatedly, sound started after next loop cycle first time, but took a couple of cycles second time; with this PR there was only one alert notification on the lock screen

Some of my questions may apply to all alerts and perhaps some revision is needed that would be a separate PR.

Questions:
* should the rec. bolus indicator remain on the snoozer screen when over threshold even though the sound has been silenced?
* when I did not snooze, the rec. bolus indicator stayed on the snooze screen after rec. bolus is no longer > threshold but the sound was never repeated after the first interval (even after I enabled repeat sound)
* I think the Snooze screen should be updated differently, but perhaps for another PR

## Test

### Test 1

Set the alert threshold to 0.5 U. The default snooze is 5 minutes.
Adjust CGM values and/or carbs on phone to make changes to rec. bolus to test feature.
Columns:
* alert means - rec. bolus shows on snoozer screen
* sound means - I heard audible alarm
* comment what I did before next row

| rec. bolus | alert | sound | comment
|:--|:-:|:-:|:--|
| 0.34 | n | n ||
| 0.54 | y | y ||
| 0.00 | n | n | add carbs but don't bolus |
| 1.21 | y | y | tap snooze |
| 0.66 | n | n | add carbs but don't bolus |
| 1.92 | y | y | tap snooze |
| 1.87 | n | n | bolus on loop phone |
| 0.00 | n | n | add carbs but don't bolus |
| 1.21 | y | y | do not snooze, noise stops after a bit |
| 0.88 | y | n | |
| 0.83 | y | n | modify alarm to have repeat sound enabled |
| - | - | - | walk away . . . |
| 0.43 | n | y | should reset |

was using a tidepool-merge build for a different test on the looping phone. That has NS issues. Got bogus Not Looping because of the tidepool-merge issue, so switch to 3.4.4 version to continue. 

### Test 2

Clear the rec. bolus alert that is still showing on Snooze screen even though rec. bolus is back to 0 by tapping snooze. The repeat sound is enabled for rec. bolus alert. Still with threshold of 0.5 U. (Note that test phone is using the bt-heartbeat feature with a RileyLink).

| rec. bolus | alert | sound | comment
|:--|:-:|:-:|:--|
| 0.00 | n | n | add carbs but don't bolus, lock LF phone |
| - | - | - | got a Rec. Bolus alert from LoopFollow on my lock screen but no sound yet, wait until next loop cycle before opening |
| 0.85 | y | n | even 9 min after lock screen alert, no sound, open LF, tap Snooze on snooze screen |
| 0.80 | n | n | add carbs but don't bolus, leave LF open |
| 2.00 | y | y | did not show initially, takes a minute or two before it sounds |

### Test 3

Build LoopFollow dev (without this PR, commit 5260225) and see if the alert sounds with a locked phone.
Use Loop phone to deliver rec bolus so we start test with rec. bolus back to 0 U.

| rec. bolus | alert | sound | comment
|:--|:-:|:-:|:--|
| - | n | n | add carbs but don't bolus, lock LF phone |
| - | - | - | got a Rec. Bolus alert from LoopFollow on my lock screen but no sound yet, wait until next loop cycle before opening |

Note - there is no sound (after 2 minute) but there were repeated Rec. Bolus notifications on the lock screen - several per minute.
Wait for next loop cycle and then there is the noise. Open LF phone and snooze the sound.

Deliver the rec bolus on Loop phone. Wait for rec. bolus to show 0.
And then add carbs again, with LF phone locked. See if sound repeats with dev branch

| rec. bolus | alert | sound | comment
|:--|:-:|:-:|:--|
| 0.00 | n | n | add carbs but don't bolus, lock LF phone |
| - | - | - | got a Rec. Bolus alert from LoopFollow on my lock screen but no sound yet, wait until next loop cycle before opening |

The intiial rec. bolus alert was about 14 (?) minutes old when the phone started to make noise.

Need to take a break - I can confirm behavior of locked LoopFollow phone with this PR later if desired.

Issue #320, opened 2024-07-20
Missing ‘+’ Delta-Sign in Notifications


bjorkert:
Description:
After the recent update, notifications on watch/phone show BG and the difference from the last reading, but do not include the ‘+’ sign. The ‘-’ sign is displayed correctly. The sign is also missing in the snooze screen.

Steps to Reproduce:
	1.	Receive a high or low BG alert on the watch/phone.
	2.	Observe the BG and the difference from the last reading in the notification.

Expected Result:
The notification should display both ‘+’ and ‘-’ signs, including for mmol/L.

Actual Result:
The ‘+’ sign is missing, while the ‘-’ sign is displayed.

Fix Required:
Include the ‘+’ sign in notifications for mmol/L values, even though this was not previously implemented.

bjorkert:
Solved by https://github.com/loopandlearn/LoopFollow/pull/367

PR #367, opened 2025-01-28
Bugfix, show delta string with +- sign in notifications

bjorkert:
Solves https://github.com/loopandlearn/LoopFollow/issues/320

The ‘+’ sign was missing in notifications, fixed in this PR.

### DIFF
--- a/LoopFollow/Controllers/Alarms.swift
+++ b/LoopFollow/Controllers/Alarms.swift
@@ -640,7 +640,7 @@ extension MainViewController {
         if !UserDefaultsRepository.alertAudioDuringPhone.value && isOnPhoneCall() { audioDuringCall = false }
         
         guard let snoozer = self.tabBarController!.viewControllers?[2] as? SnoozeViewController else { return }
-        snoozer.updateDisplayWhenTriggered(bgVal: Localizer.toDisplayUnits(String(bgData[bgData.count - 1].sgv)), directionVal: latestDirectionString ?? "", deltaVal: Localizer.toDisplayUnits(latestDeltaString) ?? "", minAgoVal: latestMinAgoString ?? "", alertLabelVal: AlarmSound.whichAlarm)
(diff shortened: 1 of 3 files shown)


marionbarker (approved):
## Code Review:

This fixes Issue #320
An optional change is offered for your consideration (remove the + from in front of 0 delta).
I do note that Nightscout uses +0 for their delta, so you may choose to be consistent with NS.

## Tests

### Test - PR as written

Observe Deltas with this PR using mg/dL, starting value of glucose was 110
```
110 +0
109 -1
111 +2
```

### Test - PR with optional modification

Switch to mmol/L display. Rebuild with optional modification.
Initial value was 111 mg/dL or 6.2 mmol/L
Show 3 numbers below: mg/dL uploaded to NS, mmol/L displayed, delta displayed
Note that the delta is correct, but user may be confused by 6.2 followed by 6.4 showing delta of +0.3
```
116 6.4 +0.3
110 6.1 -0.3
110 6.1  0.0
109 6.0 -0.1
111 6.2 +0.1
```

marionbarker:
I evidently did not include the optional change when I submitted my review.
Here's the diff I tested.

```
diff --git a/LoopFollow/Controllers/Nightscout/BGData.swift b/LoopFollow/Controllers/Nightscout/BGData.swift
index 3e0064e..ec6b3a0 100644
--- a/LoopFollow/Controllers/Nightscout/BGData.swift
+++ b/LoopFollow/Controllers/Nightscout/BGData.swift
@@ -245,7 +245,8 @@ extension MainViewController {
             // Delta handling
             if deltaBG < 0 {
                 self.latestDeltaString = Localizer.toDisplayUnits(String(deltaBG))
-
+            } else if deltaBG == 0 {
+                self.latestDeltaString = Localizer.toDisplayUnits(String(deltaBG))
             } else {
                 self.latestDeltaString = "+" + Localizer.toDisplayUnits(String(deltaBG))
             }
```

bjorkert:
Thanks for the review! I personally think keeping the + sign even before a 0 makes sense, as it clearly indicates that this value represents a delta. A standalone 0 could be confusing without that context.

Additionally, this aligns with how the delta is displayed within the app (on the main screen and snooze screen). Previously, only the notification behaved differently, so this change improves consistency across the app.

PR #370, opened 2025-02-02
Add battery drop alarm

dnovosel:
This change adds the ability to setup a battery drop alarm for situations where the battery is being consumed rapidly.

The specific use-case we ran into was based on the camera app being left opened and draining the battery rapidly without the operator knowing it was happening.

The user can setup the battery drop percentage and period of time to monitor the drop.  The implementation uses a fuzzy match to find the closest battery reading to the defined period to account for differences in polling for the battery data.

**User configured values**
- Battery Drop is between 5 - 100%
- Period (minutes) is between 5 -30 minutes
- Snooze is defined in hours

The change will store a maximum of 30 battery updates.  Which can cover anywhere from around 30 minutes to 2.5 hours depending on polling.

Possible room for improvement.  Currently the matching will find the closest battery reading, even if it somehow happens to be well outside the user defined period.  Although unlikely to be a concern in day-to-day usage, it is possible in certain scenarios that battery updates are not occurring on the devices and could trigger an alert based on bad data.  

- I could add additional logic to handle this case more graceful if we thing it's worth covering off.

### DIFF
--- a/LoopFollow/Controllers/Alarms.swift
+++ b/LoopFollow/Controllers/Alarms.swift
@@ -507,6 +507,27 @@ extension MainViewController {
             }
         }
 
+        // Check for rapid battery drop
+        if UserDefaultsRepository.alertBatteryDropActive.value && !UserDefaultsRepository.alertBatteryDropIsSnoozed.value {
+            let targetDate = Calendar.current.date(byAdding: .minute, value: Int(-UserDefaultsRepository.alertBatteryDropPeriod.value), to: Date())!
+            let currentBatteryLevel = UserDefaultsRepository.deviceBatteryLevel.value as Double
+            let dropPercentage = Double(UserDefaultsRepository.alertBatteryDropPercentage.value)
+
+            // find the closest matching entry to the user defined timeframe
+            // this allows flexibility for ingress data matching as it can come at different intervals
+            if let previousBatteryLevel = deviceBatteryData.min(by: {
+                abs($0.timestamp.timeIntervalSince(targetDate)) < abs($1.timestamp.timeIntervalSince(targetDate))
+            }) {
+                if (previousBatteryLevel.batteryLevel - currentBatteryLevel) >= dropPercentage {
+                    AlarmSound.whichAlarm = "Battery Drop"
+
+                    if UserDefaultsRepository.alertBatteryDropRepeat.value { numLoops = -1 }
+                    triggerAlarm(sound: UserDefaultsRepository.alertBatteryDropSound.value, snooozedBGReadingTime: nil, overrideVolume: UserDefaultsRepository.overrideSystemOutputVolume.value, numLoops: numLoops, snoozeTime: UserDefaultsRepository.alertBatteryDropSnoozeHours.value, snoozeIncrement: 1, audio: true)
+                    return
+                }
+            }
+        }
+
         if UserDefaultsRepository.alertRecBolusActive.value,
            !UserDefaultsRepository.alertRecBolusIsSnoozed.value
         {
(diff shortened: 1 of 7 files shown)


dnovosel:
Found a small defect I’ll patch today.  The drop calculation should ignore when the previous value was 100%.

dnovosel:
Additional observation after testing for a few more days.  The battery seems to drop in 5% increments, which makes 5% drops difficult to detect as we can get a sudden transition.

I'm going to do some additional digging to see why we see battery increments in that manner, unsure what the source is right now, as it could be Loop or Nightscout.

dsnallfot:
> Additional observation after testing for a few more days. The battery seems to drop in 5% increments, which makes 5% drops difficult to detect as we can get a sudden transition.
> 
> I'm going to do some additional digging to see why we see battery increments in that manner, unsure what the source is right now, as it could be Loop or Nightscout.

I did some googling around this, and it seems that the 5% increments is by iOS design. There are a few discussions in the dev communities that the increments have changed between different iOS versions. See this for instance: https://developer.apple.com/forums/thread/732903?page=1

The batteryLevel is fetched like this (UIKit) in Trio. Guess Loop does it in similar way. I could not find any code elsewhere in Trio or Nightscout that alters the increments fetched, so its an iOS thing as far I can tell...
Skärmavbild 2025-02-06 kl  21 06 20

bjorkert:
The code looks good overall! About the check to ignore battery drops when the previous battery reading is 100%, I’m not entirely clear on the reasoning behind that. Could you explain why 100% is treated as a special case?

marionbarker:
## Testing

Successful test

### Test with Loop phone

Use an iPhone 8 test phone with very poor battery.
Configure it to use the Loop app with rPi DASH simulator.

Use an iPhone SE phone with LoopFollow from this PR on the phone looking at the Loop URL from the iPhone 8.
* Unplug the iPhone 8 from power
* Configure the Battery Drop alarm for the LoopFollow (SE) phone
   * Alarm if battery drops by 5% over 20 minutes
   * If it alarms, snooze the alarm (for 1 hour)
* Observe iPhone 8 battery over time and observe when alarm sounds
* To make sure I don't loose track of what's going on, I also set the Battery alarm to alert at 25% battery level

| Local Time | Loop Phone Battery | LoopFollow Alarm?|
|:--|--:|:-:|
|13:47 | 100% | - |
|13:51 | 90% | - |
|14:01 | 86% | Y |

## Comments

The default alarm sound for Battery and Battery Drop are the same.
A different sound should probably be selected for the 2 different alarms.

Close out this comment for now. Let it keep running in the background - I expect I will get another alert in one hour and I'll update the table above.

marionbarker (approved):
Approve as is, but suggest the default sound for battery drop should be different from the battery alarm.

marionbarker:
### Testing continued

Continue the previous testing:

iPhone 8 tried to reboot and failed to start until it was plugged in again. 
The battery said it was at 13% when first able to see battery
Unplug again when it reached 24%
Last Battery reading in LoopFollow (and Nightscout) was 80%
Wait for next Loop cycle - have to restart the rPi simulator
Wait for an alert from LoopFollow and then terminate test.

Was too fast to hit snooze - I think this would have been the battery alarm.
Terminating tests for this PR.

## comment

It would be nice if each alarm was added to the log file. That way you can go back and see what alert alarmed after you hit snooze. But I think a complete revamp of alarms is planned, so don't expect this to happen until then.

marionbarker:
I had second thought about the sound for new Battery Drop alert
* A battery problem is a battery problem whether it is dropping fast or below a set limit.
* I agree that adding alerts records to the log is appropriately delayed for a separate PR.

LGTM

PR #372, opened 2025-02-16
Dynamic calculation of bg delay

bjorkert:
# Dynamic BG Delay

## Summary
This change implements an dynamic mechanism to calculate the delay between when a BG value is generated by the sensor and when it is available from Dexcom or Nightscout. The goal is to minimize the update delay in LoopFollow while avoiding unnecessary fetches.

## Changes
- Added dynamic delay calculation based on observed sensor-to-upload delays.
- Updated BG data processing to update the delay estimate only when a new reading is detected.
- Refactored settings to include a new "Dynamic BG Delay" option with a read-only display of the current dynamic delay.
- Moved BG delay settings into their own section, hiding the static setting when dynamic mode is enabled.
- Implemented logging for delay estimation debugging.

## Testing
Verified with real life data that the dynamic delay adapts as expected and that the settings UI reflects the current mode and delay.

### DIFF
--- a/LoopFollow/BackgroundRefresh/BT/BLEDeviceSelectionView.swift
+++ b/LoopFollow/BackgroundRefresh/BT/BLEDeviceSelectionView.swift
@@ -13,13 +13,14 @@ struct BLEDeviceSelectionView: View {
     var body: some View {
         VStack {
             List {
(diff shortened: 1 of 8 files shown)


bjorkert:
Example:
CleanShot 2025-02-17 at 09 17 22

Issue #371, opened 2025-02-16
2.2.9 isn’t keeping prediction lines in sync


aidanlane:
Hey team, loving your work on Loop Follow!

2.2.9 isn’t keeping prediction lines in sync with BG readings and treatments.

Using Trio 0.2.3 with Nightscout hosted by Nightscout Pro.

Using a Dexcom G6 with OmniPod Dash.

Running on an iPhone 15 with iOS 18.3.1.
Using the silent audio background refresh.

Reverting to Loop Follow 2.2.8 worked a charm. Tried switching back to 2.2.9 again and had the issue again. Then back to 2.2.8 and it worked again. So easily reproducible, at least for me 😉

See the attached screenshots.

Happy to send any logs or try anything to help debug it.

aidanlane:
![Image](https://github.com/user-attachments/assets/58276ca6-bf44-4580-bbe9-f30553b71e7a)
![Image](https://github.com/user-attachments/assets/46cd9125-ea72-48bf-977c-181282f7b296)
![Image](https://github.com/user-attachments/assets/04abf42e-7ee7-4c95-830a-c5edbe7553a7)
![Image](https://github.com/user-attachments/assets/76c81ac3-bb5c-4957-b8e7-f48d876ff112)

aidanlane:
For reference, on 2.2.8:

![Image](https://github.com/user-attachments/assets/78d9f0ef-7a21-4ac8-9da6-45bf2d26c6af)

marionbarker:
I confirmed your error.

Running LoopFollow v2.2.9:
* view a URL from Trio 0.2.3 - no prediction lines
* using URL uploaded to Nightscout from test phone
   * when test phone is running Trio closed beta - prediction lines are displayed
   * build Trio 0.2.3 on that same test phone - prediction lines are not displayed on LoopFollow v2.2.9

Revert to LoopFollow v2.2.8
* test URL from Trio 0.2.3, Trio closed beta and Loop - all display prediction lines as expected.

Apologies - for not testing thoroughly enough before we merged this.
We will find the issue and update.

aidanlane:
You’re amazing, thanks @marionbarker!

marionbarker:
Thanks again for reporting this. @bjorkert provided the fix in PR #373 and I released an update.
Please try again with LoopFollow v2.2.10.

Closing this issue.

aidanlane:
[...]

PR #373, opened 2025-02-21
Predictions for released version of Trio

marionbarker:
Fix for a bug introduced in v2.2.9.
This restores proper display of predictions for released version of Trio, Trio 0.2.3 and earlier.
See Issue #371 for details.


marionbarker:
## Status

LGTM

## Tested

This modification restores the display of prediction lines for Trio 0.2.3
* let this run with a Trio 0.2.3 URL for about half-an-hour and observe the predictions are as expected.
* test against a Trio-dev URL and a Loop URL to confirm no change to those applications, predictions are displayed as expected

bjorkert (approved):

PR #375, opened 2025-02-23
Estimated bg delay

bjorkert:
## Overview

This PR introduces a new feature for Dexcom devices that calculates the sensor’s time offset from midnight (in 5-minute cycles) and compares it with the device’s heartbeat offset. By incorporating a polling delay (typically 9–10 seconds) — which is the time needed for BG values to become available in Nightscout — the system can display the expected delay for each sensor available as a heartbeat device.

telegram-cloud-photo-size-4-5897815023642854756-y
The screenshot demonstrates the expected BG delay calculation for different Dexcom devices. 

- The **selected device (DXCMnw)** shows an expected delay of **274 sec**.
- Another detected device (**DXCMGP**) has a slightly lower delay of **247 sec**.


marionbarker:
## Status

LGTM

## Test

Requested a modification to the hint - add explanation for how to use the added information of expected bg delay.

Code with this modification presents the choices (to me) in the graphic below for Dexcom. (I have 6 used G7 sensors in the room but I must also have a G6 and another G7 nearby (the lower signal indicates either older or farther away.) I indicated my choice (balance greatest signal strength and shortest estimated delay).

lf_pr375_dexcom

marionbarker (approved):

PR #378, opened 2025-03-03
Omnipod DASH heartbeat

bjorkert:


marionbarker (approved):
I added a commit to enable use of an rPi DASH simulator when using OmnipodDASH for Background Refresh Type.

Warning-if using an rPi DASH in the same room as an rPi Dana simulator, the dana bluetooth name shows up in preference to the " :: Fake POD ::" name.

PR #382, opened 2025-03-08
Auto build and nuke certs

bjorkert:


marionbarker:
I started a fresh branch and believe we should use PR #383 instead.

PR #384, opened 2025-03-11
Task Scheduler Thread Safety Improvements

bjorkert:
## Overview
This PR enhances the thread safety of the Task Scheduler by replacing the main-thread `Timer` with a `DispatchSourceTimer`. All timer scheduling and task management now occur on a dedicated serial queue, eliminating potential data races even when public methods are invoked from arbitrary threads.

## Changes
- **DispatchSourceTimer Implementation:**  
  - Replace `Timer` with a `DispatchSourceTimer` scheduled on the Task Scheduler’s serial queue.
- **Safe Publication:**  
  - Configure the timer in a local variable before assigning it to `currentTimer`, ensuring the timer is fully configured before exposure.
- **Thread Safety:**  
  - Confine all accesses to shared state (`tasks` dictionary and `currentTimer`) to a single serial queue.
- **Refactoring:**  
  - Minor code cleanup and improved logging for better traceability.

## Testing
- Verified that scheduled tasks execute as expected.
- Confirmed that thread sanitizer warnings and data race conditions are resolved.
- Conducted both unit tests and manual testing to ensure overall stability.

### DIFF
--- a/LoopFollow/Task/TaskScheduler.swift
+++ b/LoopFollow/Task/TaskScheduler.swift
@@ -27,11 +27,10 @@ struct ScheduledTask {
 class TaskScheduler {
     static let shared = TaskScheduler()
 
-    // Thread-safety: a serial queue so we don’t manipulate tasks from multiple threads at once
     private let queue = DispatchQueue(label: "com.LoopFollow.TaskSchedulerQueue")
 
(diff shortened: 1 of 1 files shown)


marionbarker (approved):
Not an expert regarding the code review.

Tested that it builds and works with the supported URL sources.

PR #389, opened 2025-04-13
Log app version and build details

bjorkert:
Enhanced the logging to include detailed build information. When the app starts and for when a new log file is created it now logs the app version, build date, branch info, and, if not running on a Mac or Simulator, the expiration date.

### DIFF
--- a/LoopFollow/Helpers/BuildDetails.swift
+++ b/LoopFollow/Helpers/BuildDetails.swift
@@ -51,7 +51,25 @@ class BuildDetails {
         return "sandboxReceipt".caseInsensitiveCompare(receiptName) == .orderedSame
 #endif
     }
-        
+
+    // Determine if the build is for Simulator
+    func isSimulatorBuild() -> Bool {
+#if targetEnvironment(simulator)
+        return true
+#else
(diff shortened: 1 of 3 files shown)


marionbarker:
### Status

I like the detailed build information in the log, so I am approving this PR. 

### Tests

I built it to real phones, simulator and the Mac.

IMG_6776

### Optional modification

I think the amount of logging needs to be reduced. You might choose to do this in another PR. For example, the Task Scheduler information is pretty verbose.

See sample logs below.

This example is on a test phone looking at a URL from another test phone which is not currently Looping.

[LoopFollow 2025-04-13-test-phone.log](https://github.com/user-attachments/files/19726940/LoopFollow.2025-04-13-test-phone.log)

This example is from a real phone viewing the URL for a running Loop instance.

[LoopFollow 2025-04-13.log](https://github.com/user-attachments/files/19726990/LoopFollow.2025-04-13.log)

marionbarker (approved):

bjorkert:
Agree about Task Scheduler logging, I need to think about that. Merging this one meanwhile.

Issue #387, opened 2025-03-16
Logs for Remote functions in Trio


tredair7:
Would it be possible to add log entry when a Remote command is issued to Trio?

bjorkert:
Included in release https://github.com/loopandlearn/LoopFollow/releases/tag/v2.3.1

PR #390, opened 2025-04-14
Add logging for remote commands

bjorkert:
Adds logging for remote commands.
image

Solves: https://github.com/loopandlearn/LoopFollow/issues/387


marionbarker (approved):
## Code Review

Looks good.

## Test

Tested with a Trio-dev URL and LoopFollow set up for Trio Remote Control.

Issued various remote commands from the LF phone. These were all received and enacted by the Trio phone and the [APNS] events were recorded. in the log.

PR #391, opened 2025-04-14
Ignore old enacted timestamps

bjorkert:
Sometimes an old enacted is uploaded, older than the previous one. This check ignores the enacted timestamp update if it is older than the last one on record.


marionbarker (approved):
## Code Review

Approved based on code review

## Test

Can't test the stale enacted dates, but make sure it builds and continues to run with a Trio-dev URL.

PR #395, opened 2025-04-15
Remove Trio limitation

bjorkert:
Currently, Trio Remote Commands are limited to dev (actually non-main releases). With this PR, they are available in main as well. This PR should be merged and released when Trio dev goes public.

### DIFF
--- a/LoopFollow/Remote/Settings/RemoteSettingsView.swift
+++ b/LoopFollow/Remote/Settings/RemoteSettingsView.swift
@@ -31,23 +31,15 @@ struct RemoteSettingsView: View {
 
                     remoteTypeRow(type: .nightscout, label: "Nightscout", isEnabled: true)
 
-                    if BuildDetails.default.branch?.lowercased() != "main" {
-                        remoteTypeRow(
-                            type: .trc,
-                            label: "Trio Remote Control",
-                            isEnabled: viewModel.isTrioDevice
-                        )
-                    }
+                    remoteTypeRow(
+                        type: .trc,
+                        label: "Trio Remote Control",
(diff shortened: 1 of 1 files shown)


LoopFollow/Remote/Settings/RemoteSettingsView.swift
```diff
@@ -31,23 +31,15 @@ struct RemoteSettingsView: View {
...
-                            isEnabled: viewModel.isTrioDevice
-                        )
-                    }
+                    remoteTypeRow(
+                        type: .trc,
+                        label: "Trio Remote Control",
+                        isEnabled: viewModel.isTrioDevice
+                    )
 
                     Text("Nightscout is the only option for the released version of Trio and Loop.")
```
marionbarker, line 40: Please remove "Trio and" from this line.

bjorkert: Thanks, text updated.

marionbarker (changes requested):
I think the text should be updated more. 

The graphic below shows the screen when a Loop URL is selected (left) and a Trio-dev URL is selected (right). I would indent the subtext and have one subtext under Nightscout for "Loop or Trio 0.2.x" and another subtext under Trio Remote Control that says something like "Trio 0.4.x or later".

lf-remote-loop-trio

Other problem, I set up a Trio 0.2.x URL and LoopFollow allows selection of Trio Remote Control but it is not supported with that version of Trio. Not sure if there is a way to know in LF if the Trio version really supports Trio Remote Control. I assume not, in which case the text update will have to be sufficient.

marionbarker (approved):

PR #396, opened 2025-04-22
Resolve conflict

bcdiabetes:


.github/workflows/build_LoopFollow.yml
```diff
@@ -1,16 +1,11 @@
 name: 4. Build Loop Follow
 run-name: Build Loop Follow (${{ github.ref_name }})
 on:
+  schedule:
+    # Runs "At 11:00 on every day-of-week from Monday through Friday"
```
MikePlante1, line 5: This comment does not match the line below it, which runs at 17:00 UTC on the 20th day of every month.

PR #397, opened 2025-04-22
disable keepalive-workflow because the repo is no longer available

marionbarker:
## Purpose

The `gautamkrishnar/keepalive-workflow` has been taken down by GitHub. ~~We do not know if this is permanent or not.~~ We now know this is a permanent change.
This PR makes a minimal change to the current build_LoopFollow.yml which renables building with GitHub actions.

## Method

Comment out the section that is not available and insert an echo for the log file.

At a later time, the whole logic thread for automatic build will be reviewed, updated and documented.

## Testing

This was copied from the LoopWorkspace change and requires test.


bjorkert (approved):
Tested: https://github.com/bjorkert/LoopFollow/actions/runs/14602859088

bjornoleh (approved):

PR #400, opened 2025-05-19
Fix for crashes related to overrides

bjorkert:
What’s fixed
Nightscout override segments are now sanitised before they hit the graph layer, so they never overlap and are never shorter than 5 minutes.

Why it crashed before
I trimmed each segment while iterating newest → oldest.
That meant I sometimes pushed the segment’s end earlier than its start, giving negative (or zero-length) durations.
The graph code asserts that duration > 0, so the app crashed as soon as one of those bad dots was drawn.

How it’s solved
	1.	Sort overrides oldest → newest first.
	2.	For each segment, cap its end at (nextStart − 5 min).
	3.	Skip any segment that ends up < 5 min long.

The result is a clean, strictly sequential timeline, so the graph renders without blowing up.

### DIFF
--- a/LoopFollow/Controllers/Nightscout/Treatments/Overrides.swift
+++ b/LoopFollow/Controllers/Nightscout/Treatments/Overrides.swift
@@ -10,79 +10,80 @@ import Foundation
 import UIKit
 
 extension MainViewController {
-    // NS Override Response Processor
-    func processNSOverrides(entries: [[String:AnyObject]]) {
+
+    func processNSOverrides(entries: [[String: AnyObject]]) {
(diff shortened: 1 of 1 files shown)


marionbarker:
## Test

Use a URL (provided privately) to demonstrate the crash.

Use PR branch. It fixed the crash but the resultant gaps in the override plot, where overlaps were truncated, were large (5 min each). Modify this to be 1 minute instead. Add comments to make it clear where overlaps are being removed and short overrides are being skipped.

marionbarker (approved):
Reviewed and tested. LGTM.

PR #401, opened 2025-05-20
Add auto bump of dev version on push

marionbarker:
## Purpose:

Differentiate different code by version numbers.
For each release, main and dev start out being identical, e.g., both at Major.Minor.0

Each time a PR is added to the dev branch, modify the version by incrementing the 3 integer.

## Questions

* Not sure how this will work with a release - does dev get incremented when it is still the same as main.
* Not sure if we want this level of automation added to LoopFollow
* I did some testing (with a different branch) in LoopFollow_Third, but I don't have the fine-grained PAT configured yet

### DIFF
--- a/.github/workflows/auto_version_dev.yml
+++ b/.github/workflows/auto_version_dev.yml
@@ -1,0 +1,91 @@
+# -----------------------------------------------------------------------------
+# Workflow: Bump Dev Version
+#
+# Description:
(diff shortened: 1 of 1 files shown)


dnzxy (approved):
Code looks good.
We were zooming, when you did most of the setup, so I know this is good.
Successful workflow runs for confirmation.

LGTM ✅

PR #402, opened 2025-05-23
Fix for Loop indefinite overrides

bjorkert:
This PR fixes a regression introduced in https://github.com/loopandlearn/LoopFollow/pull/400, where indefinite overrides created by Loop were incorrectly displayed as 5-minute events until they ended. Once terminated, they were updated with the correct duration and rendered properly.

With this fix, indefinite overrides are now shown correctly from the start.

### DIFF
--- a/LoopFollow/Controllers/Nightscout/Treatments/Overrides.swift
+++ b/LoopFollow/Controllers/Nightscout/Treatments/Overrides.swift
@@ -38,7 +38,13 @@ extension MainViewController {
             else { continue }
 
             let start = max(startDate.timeIntervalSince1970, graphHorizon)
-            var end   = start + (e["duration"] as? Double ?? 5) * 60   // seconds
+
+            var end: TimeInterval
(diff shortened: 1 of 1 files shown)


marionbarker:
## Test

### Confirm the problem

Configure a test phone to upload to NS and be followed using LoopFollow 2.5.0
* Issue an indefinite override on the Loop phone
* Wait at least one loop cycle for NS to update
   * Observe that the indefinite override shows up as 5 minutes  extent while running
* Cancel the override on the Loop phone 
   * Observe that after canceling the override, the extent shown is the actual time it was engaged

### Confirm this PR works as expected

Configure a test phone to upload to NS and be followed using this branch
* Issue an indefinite override on the Loop phone
* Wait at least one loop cycle for NS to update
   * Observe that the indefinite override shows up indefinite - watch for a few more cycles
* Cancel the override on the Loop phone after 42 minutes
   * Observe that after canceling the override, the extent shown is the actual time it was engaged

PR #408, opened 2025-06-12
Improved Alarm system, Snoozed screen, Landscape and more

bjorkert:
High level list of changes
* Alarms are now more customizable, you can add and remove alarms, you can have several alarms of the same sort with different thresholds. Existing alarms are migrated.
* New alarm, Looping App Expire (currently only for Trio)
* Some alarms can now be acknowledged instead of snoozed (snooze 0 minutes)
* Shorter (or no) delay before alarms are triggered after the bg value is available
* Landscape mode support
* Link to the LoopFollow Facebook group
* Snoozer screen now has an optional Emoji illustrating the current bg value
* Remote settings has hidden information in order to hide secrets in screenshots
* Lots of under the hood refactoring, cleanup and modernization
* The Eureka framework is now 100% replaced by Swift UI
* Introduced swiftformat for code formatting and header information


marionbarker:
## Initial Test

When I add an alarm, it shows up on my alarms page and it stays on the list until I delete the alarm.
* It is nice to be able to just show the alarms I want to have handy

The indication that a particular alarm is enabled or is not subtle - not sure if we want to be more emphatic:

loopfollow-alarms-pr408


## Possible improvements:

I'm not sure if folks would 
* prefer alarms to sort so the enabled ones are at the top
* like to be able to move alarms up and down on the screen
* like to have a quick view if alarms have fired and are snoozed

marionbarker (approved):
Approved from cursory test.

* I am not going to review the code on this - it would probably be good if someone did.
* The quick test looks like the alarms work - but also, not going to test every instance right now.

Should we get some actual users to build this and test it out?

bjornoleh:
I have built and briefly started testing. Looks good so far, many nice improvements!

Sorting by active alarms helped a lot to make the active ones stand out more, I built both before and after that last change.

Issue #404, opened 2025-05-27
Trio Remote Control - Temp Target screen timeout


woodduckie:
## Describe the bug
*When creating a new temp target, either to enact directly or save as preset, Loop Follow will revert to the "Trio Remote Control" screen without input, warning, or message*

## To Reproduce
*Steps to reproduce the behavior:*
1. *Go to "Remote" icon*
2. *Click on "Temp Target" icon*
3. *Enter target and duration values*
4. *Wait for up to ~30 seconds.  Sometimes so short that I cannot select "Save as Preset" in time to retain.*
5. *Loop Follow reverts to the "Trio Remote Control" screen without prompting*

## Expected behavior
*Allow for a longer period of time for user to enact or save as preset - say 60 seconds minimum.*

## Setup Information:
*Using Remote Settings: "Trio Remote Control"*

### Smartphone:
* Hardware: *iPhone 15 Pro Max*
* OS Version: *iOS 18.5*

### Loop Follow Version:
* Version Number: *2.5.0*
* Repo: *loopandlearn / LoopFollow*
* Git Reference: *Main 3392554*

## Additional context
*When enacted in time, remote settings are effective and functional.  This does not appear to be an error with authentication settings.*

PR #412, opened 2025-06-18
RemoteViewController: suppress duplicate @Published emissions to stop…

bjorkert:
### What does this PR do?
Rebuilds of **`RemoteViewController`** were triggered every time `remoteType` or `device` was **re-assigned**—even when the value didn’t actually change—because `@Published` emits on every write.  
That tore down the `NavigationView` and pushed users back to the root screen, so Bolus/Meal/Temp Target windows closed unexpectedly.

### Fix
Adds `.removeDuplicates()` to both publishers in the `CombineLatest` chain.  
Identical values are now filtered out, so `updateView()` runs only when the user genuinely switches Remote Type or Device.

### User-visible impact
* Bolus, Meal, Temp Target, and Overrides screens stay open; no more spontaneous “pop”.
* No change in behaviour when the user **actually** changes the setting.

### Reference / evidence
* Facebook report by **Muntean George** (Trio: DIY Diabetes App Support group).  
* GitHub issue **#404** “Trio Remote Control – Temp Target screen timeout”.

### Tested
1. Open Bolus screen and wait 10 minutes → stays on screen.  
2. Change Remote Type in Settings → screen refreshes as before.

### DIFF
--- a/LoopFollow/Remote/RemoteViewController.swift
+++ b/LoopFollow/Remote/RemoteViewController.swift
@@ -21,8 +21,8 @@ class RemoteViewController: UIViewController {
         super.viewDidLoad()
 
         cancellable = Publishers.CombineLatest(
-            Storage.shared.remoteType.$value,
-            ObservableUserDefaults.shared.device.$value
+            Storage.shared.remoteType.$value.removeDuplicates(),
+            ObservableUserDefaults.shared.device.$value.removeDuplicates()
         )
         .sink { [weak self] newRemoteType, newDevice in
             DispatchQueue.main.async {


marionbarker:
## Test

### Confirm the error

Build LoopFollow v2.8.1 onto a test phone that is observing a test Trio URL.
* Configure the phone for remote control
* Test that meal entry is sent as expected
* Open the LoopFollow, Remote, Bolus screen and wait while doing other things
* In about 5 minutes, it popped back to the nominal remote screen

### Confirm the fix

Build LoopFollow with the branch onto a test phone that is observing a test Trio URL.
* Test that meal entry is sent as expected
* Open the LoopFollow, Remote, Bolus screen and wait while doing other things
* More than 10 minutes later - Bolus screen is still open
* successfully send remote bolus

PR #415, opened 2025-06-28
Add more comments at pauses for release.sh [ci skip]

marionbarker:
After using the release.sh for the first time, decide to add some more comments at the pauses and update some of the comments that are already there.


bjorkert (approved):
LGTM

Issue #405, opened 2025-06-01
Trio LF-remote: Note regarding rejected bolus is not readable


bjornoleh:
The note from Trio regarding rejected remote bolus is not fully readable. The text is long and displayed as a single line, and even scrolling sideways will not reveal the start and end of the note. 

It would be nice (and expected) to be able to read this in NS after attempting a remote bolus.

The text is very long, not sure how it can be displayed in full, but multi line wrapping is perhaps possible?

<img src="https://github.com/user-attachments/assets/01ae3388-2c50-49ad-8395-5a7b3ba72384" width="300">

Full text from NS:

`Trio	

Command rejected: boluses totaling more than 20% of the requested amount have been delivered since the command was sent. Details: User: X. Command Type: Bolus. Bolus Amount: 0.29999999999999999 units `

marionbarker:
Fixed with PR #416 which is merged into dev. This fix will be in the next LoopFollow release.

Thank you for your report.

PR #416, opened 2025-07-01
Wrap note texts

bjorkert:
This PR solves the issue where texts are to long to fit on the screen, as described in https://github.com/loopandlearn/LoopFollow/issues/405

Now note texts are wrapped at 40 characters.

CleanShot 2025-07-01 at 21 04 11@2x


marionbarker (approved):
approve from code review and test

Issue #333, opened 2024-10-28
Feature Request: Select buttons in toolbar


marionbarker:
Some prefer to have alarms in the toolbar slot that was changed to Remote Control in v2.2.5

This is especially true for Loop users who cannot use the Remote Control features in LoopFollow.
There may be other users who primarily use LoopFollow for Alarms and having alarms be in the toolbar is convenient for them.

**Possible Idea:**
Choose options for the 3 middle buttons? Perhaps from this set with Home and Settings remaining fixed on either side of the toolbar?
* Remote
* Alarms
* Snoozer
* Nightscout

dsnallfot:
Just FYI how this could be approached. I've experimented a bit with this in a custom build of mine. It seems like it is UIKit default behavior to create a "More ..." tab on the right side when you add more than 5 tabs in the tab bar, which also allows users to drag and drop what tabs that they want to display in the tab bar, and which should be "hidden" behind the more-tab.

You can configure which tabs that should be possible to organize by the user - In my example below "Remote" and "Nightscout" is user drag n drop-able (other tabs are in fixed positions). The latest selection under the more-button also shows up as default when you tap more.


https://github.com/user-attachments/assets/6b22df14-c92e-4e7a-9231-c43c98d61c5b

bjorkert:
This issue has been resolved with the release of LoopFollow 3.1
https://github.com/loopandlearn/LoopFollow/releases/tag/v3.1.0

PR #420, opened 2025-07-03
Make Second Tab Configurable

bjorkert:
### **Description**

This PR introduces a dynamic second tab in the tab bar. 🛠️

For users who have remote functionality **enabled** (i.e., `remoteType` is not `.none`), the second tab will continue to show the **Remote** view.

For users who do **not** use the remote features (`remoteType` is `.none`), this change restores the **Alarms** view to the second tab. This provides these users with quick and convenient access to alarm configuration without having to navigate through the settings menu.

This implementation listens for changes to the `remoteType` setting and updates the tab bar in real-time.


marionbarker (approved):
## Limited addition

This only covers part the request (from Issue #333) but I understand that expanded capability will need to wait.

This code change enables users who do not use remote control to have the alarm button show up on the task bar. It limits users to either remote control (if enabled) or alarm button (only if remote control is disabled) to show up in the second spot of the task bar.

## Code Review

Code changes appear appropriate for the limited change provided.

## Test

The alarm button shows up in the second spot of the task bar when remote control is set to none.

marionbarker (changes requested):
Oops - there is a linting change that was not committed. Might want to fix that before merging.

LoopFollow/ViewControllers/AlarmViewController.swift

bjorkert:
Thanks, i re-built and pushed the result of the Lint.

Issue #343, opened 2024-12-12
Speak BG repetition after Remote commands


Mia021:
2.2.7 version - when enabling or disabling remote override while on speak BG alarm - it always every time repeats the numbers

PR #421, opened 2025-07-04
Restore Speak BG for Bluetooth and Improve Announcement Logic

bjorkert:
### Description

This PR addresses two key issues with the "Speak BG" feature:

1.  It restores functionality for users on the **Bluetooth** background refresh mode.
2.  It fundamentally improves the announcement logic to prevent incorrect repetitions.

#### Restoring Bluetooth Background Mode

The "Speak BG" feature was failing for users on the Bluetooth background refresh setting because iOS was denying the app's request to activate an audio session from a background state.

This has been fixed by implementing a more robust `AVAudioSession` management strategy:
* The audio session is now activated only when a new BG reading is ready to be announced.
* The session is intelligently deactivated immediately after speech is finished, but **only when necessary**. It correctly keeps the session alive when in "Silent Tune" background mode but closes it for Bluetooth mode to save power.

#### Improved Announcement Logic & Bug Fix

The previous logic used a simple 30-second timer to prevent repeated announcements. This was inefficient and led to bugs where old BG values could be repeated after certain actions, like remote commands.

The announcement logic has been completely rewritten:
* Instead of a timer, the app now tracks the **timestamp** of the last BG value that was announced.
* It will only speak a new value if its timestamp is more recent than the last one spoken.

This new data-driven approach is more reliable and naturally prevents the repetition of old data, which directly resolves the bug reported in issue #343.

---
**Resolves:** #343

### DIFF
--- a/LoopFollow/Controllers/SpeakBG.swift
+++ b/LoopFollow/Controllers/SpeakBG.swift
@@ -121,37 +121,38 @@ extension MainViewController {
     }
 
     // Speaks the current blood glucose value and the change from the previous value.
-    // Repeated calls to the function within 30 seconds are prevented.
     func speakBG(currentValue: Int, previousValue: Int) {
-        let audioSession = AVAudioSession.sharedInstance()
-        do {
-            try audioSession.setCategory(.playback, mode: .default)
-            try audioSession.setActive(true)
-        } catch {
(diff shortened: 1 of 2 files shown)


marionbarker (approved):
## Code review

Code review looks appropriate

## Test

I have not tested this by reproducing the error and demonstrating that it is fixed.
I have built it and heard speakBG work as expected while app is in the foreground.

Issue #403, opened 2025-05-25
The override text in the info table is glitchy


bjorkert:
At least in trio the override name is not displayed at all times.

bjorkert:
This issue has been resolved and is included in the 3.0 release. Closing.

PR #423, opened 2025-07-04
Clear override text conditionally based on device

bjorkert:
This pull request resolves issue #403.

**Description:**

This change introduces a conditional clearing of the override information in the `infoManager`. Previously, the override data was cleared every time the device status was updated.

This caused an issue for Trio users, where the override text would disappear from the display. This happened because for Trio, override information is fetched from treatments, and the device status does not contain this data. If a device status update occurred after the treatment data was processed, the override information would be incorrectly cleared.

For "Loop" devices, the override information *is* part of the device status, so clearing it upon refresh is the correct behavior to ensure the display is up-to-date.

**Changes:**

- In `updateDeviceStatusDisplay(jsonDeviceStatus:)`, the call to `infoManager.clearInfoData(types: [.override])` is now wrapped in a conditional check.
- The override data is now only cleared if the selected device is "Loop".

This ensures that the override information is handled correctly for both Loop and Trio users, fixing the glitch where the override text would intermittently disappear for the latter.

### DIFF
--- a/LoopFollow/Controllers/Nightscout/DeviceStatus.swift
+++ b/LoopFollow/Controllers/Nightscout/DeviceStatus.swift
@@ -72,7 +72,7 @@ extension MainViewController {
 
     // NS Device Status Response Processor
     func updateDeviceStatusDisplay(jsonDeviceStatus: [[String: AnyObject]]) {
-        infoManager.clearInfoData(types: [.iob, .cob, .override, .battery, .pump, .target, .isf, .carbRatio, .updated, .recBolus, .tdd])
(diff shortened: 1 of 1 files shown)


marionbarker (approved):
## Code review

The code review looks appropriate

## Test

I built this branch (as well as dev branch) on a test phone viewing both a Loop URL and  Trio URL (at separate times). I did not notice a difference with the Loop URL between the `dev` and `trio-override` builds.

**Trio URL**
* With `dev`:
   * disabling an override removed the override from the display completely (instead of changing the end point)
* With `trio-override`:
   * disabling an override updated the end point to the actual duration

Issue #422, opened 2025-07-04
Refine 'Not Looping' Alarm Logic to Consider Data Staleness


bjorkert:
#### Current Behavior
When LoopFollow is launched after being terminated for a significant period, it can sometimes trigger a false "Not Looping" alarm immediately on startup.

#### The Problem
This happens because the alarm logic evaluates the timestamp of the last successful loop. If the app was killed, this timestamp is old by default. The current logic doesn't account for the fact that the data itself is **stale** and the app hasn't had a chance to perform a fresh fetch to get the current loop status. An alarm is therefore triggered based on outdated information.

#### Suggested Solution
The validation logic for the "Not Looping" alarm should be enhanced to prevent these false positives.

When checking if an alarm should be triggered, the logic should verify two conditions:
1.  That the last recorded loop time is older than the configured threshold (current logic).
2.  **That the data itself is recent.** The app should confirm that the last *check* for the loop status happened recently enough to be considered valid.

An alarm for "Not Looping" should only be triggered if we have **recently confirmed** that the loop is not running. On a fresh app start, the app should prioritize fetching new data before evaluating alarm conditions based on the now-stale, previously stored data.

PR #424, opened 2025-07-04
Prevent stale "Not Looping" alarms

bjorkert:
### Fix: Prevent stale "Not Looping" alarms

This pull request resolves issue #422 by preventing the "Not Looping" alarm from firing when the app itself has not been able to recently check for a new status from Nightscout.

#### The Problem

The "Not Looping" alert could persist incorrectly if the app hasn't been active for a while, since the `lastLoopTime` was not being updated, it would trigger the "Not Looping" alarm, even though the issue was that the app did not fetch the data, not the loop itself.

#### The Solution 🛡️

To fix this, I've introduced a new timestamp, `lastLoopingChecked`, which is updated every time the app *attempts* to fetch the device status.

The `NotLoopingCondition` now includes a new safeguard:
1.  It checks when the last status fetch was *initiated*.
2.  If the last check attempt is more than 6 minutes old, the alarm condition is ignored.

This ensures that the "Not Looping" alarm only triggers when the app tries communicating with Nightscout rather than firing a false alarm due to stale data.

### DIFF
--- a/LoopFollow/Alarm/AlarmCondition/NotLoopingCondition.swift
+++ b/LoopFollow/Alarm/AlarmCondition/NotLoopingCondition.swift
@@ -8,7 +8,7 @@ struct NotLoopingCondition: AlarmCondition {
     static let type: AlarmType = .notLooping
     init() {}
 
-    func evaluate(alarm: Alarm, data: AlarmData, now _: Date) -> Bool {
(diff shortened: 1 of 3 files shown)


marionbarker (approved):
## Code review

Approve from code review.

PR #429, opened 2025-07-07
Reset Settings Navigation on Tab Switch

bjorkert:
## Reset Settings Navigation on Tab Switch

**Problem**: When switching between tabs and returning to Settings, the navigation stack remained at the last viewed screen instead of returning to the root settings menu.

**Solution**: 
- Added `settingsPath` to the shared `Observable` class to manage settings navigation state
- Updated `SettingsMenuView` to use the centralized navigation path
- Reset navigation path in `SettingsViewController.viewWillAppear()` to ensure users always return to the root settings screen

**Benefits**:
- Consistent user experience when switching tabs
- Follows existing app architecture patterns
- Centralized navigation state management

### DIFF
--- a/LoopFollow/Settings/SettingsMenuView.swift
+++ b/LoopFollow/Settings/SettingsMenuView.swift
@@ -9,18 +9,17 @@ struct SettingsMenuView: View {
     // MARK: - Observed Objects
 
     @ObservedObject private var nightscoutURL = Storage.shared.url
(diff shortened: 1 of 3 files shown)


marionbarker (approved):
Approving from test.

## Code review

The code review looks reasonable. 

## Test

After building with this branch, tapping the settings icon on the toolbar takes the user to the top level settings screen.

This is the desired behavior.

PR #431, opened 2025-07-08
Fix Nightscout token autofill by adding textContentType

bjorkert:
### Problem
When users select and delete the Nightscout URL, they have to navigate back to settings and return to the NS URL view before the password autofill feature works again for the token field.

### Solution
- Added `textContentType` parameter to `TogglableSecureInput` component
- Set `textContentType: .password` for the Nightscout token field
- This properly identifies the field to iOS's autofill system, maintaining consistent behavior even after URL changes

### Changes
- Modified `TogglableSecureInput.swift` to accept optional `textContentType` parameter
- Updated `NightscoutSettingsView.swift` to specify `.password` content type for token field
- Applied content type to both secure and visible states of the input

### Testing
- Verified that password autofill now works consistently after URL deletion
- Confirmed no regression in existing functionality

### DIFF
--- a/LoopFollow/Helpers/Views/TogglableSecureInput.swift
+++ b/LoopFollow/Helpers/Views/TogglableSecureInput.swift
@@ -10,6 +10,7 @@ struct TogglableSecureInput: View {
     let placeholder: String
     @Binding var text: String
     let style: Style
+    var textContentType: UITextContentType? = nil
 
     @State private var isVisible = false
     @FocusState private var isFocused: Bool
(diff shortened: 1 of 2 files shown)


marionbarker:
After applying the modification suggested by @bjorkert, commit aa9252e, this PR is a definite improvement.

Once the password icon is triggered to show up, the cursor can be on either the URL line or the token line and both URL and token are updated from the selection of a saved password.

Note - the saved passwords have the NS URL as the username and the token as the password. I have a series of these passwords I use to test all the types of URLs (Trio, Loop, Loop with no treatments) when testing LoopFollow updates.

PR #434, opened 2025-07-09
Add remote commands via APNS for Loop users

codebymini:


marionbarker:
The app stops responding when modifying various rows in the Configure Loop APNS Settings section.

Enclosed is the most recent ips file from my iPhone 15 pro.

[Loop Follow-2025-07-17-154247.ips.zip](https://github.com/user-attachments/files/21309197/Loop.Follow-2025-07-17-154247.ips.zip)

Once configured, the remote commands worked well. I'll do more testing after the crash is fixed and the branch conflicts are resolved.

marionbarker:
I tested again. (I used the latest commit, d593cce, merged dev to it and resolved the conflict).

I can select all and remove each of the items in:
* LOOP APNS KEY ID
* LOOP DEVELOPER TEAM ID
* LOOP APNS KEY

I can then add each item back in.

I can use the Camera to rescan the QR code URL if the row is already filled out.
* If I clear that row - the app hangs, when I restart, it is cleared
* If the row is cleared and I use the camera to scan, the app hangs but when I restart, the row is filled in

With this crash I am getting ips files on the SE 2nd gen but not on the iPhone 15 pro, even though they both act the same. Here's one:
* [Loop Follow-2025-07-18-102641.ips.zip](https://github.com/user-attachments/files/21321206/Loop.Follow-2025-07-18-102641.ips.zip)

marionbarker:
## Test Summary

If there is a network error - I think message should be network error - check before sending again and return to main screen

Is there a way to make sure the NS URL and the TOTP are from the same site? And if they don't match - warn the user?

## Test Details

This is with SE 2nd gen phone running iOS 18.5:

I have my APNS credential entered.
I have NS test site 3 entered as URL (Loop site)

As an experiment, I changed the QR Code URL to point to a different site and confirmed the value updated

Then I issued an override
* the overrides shown are from site 3 (not the new site)
* the override selected went to site 3 (not the new site)
* the NS URL continued to point to test site 3
* remote command for bolus and carb claimed success on LoopFollow phone but stopped working on Loop phone
* I think this is OK: Overrides work without need for the TOTP, only bolus and carbs need that
* But it would be nice if the status ok went to red if the NS URL and the QR don't come from the same site.

Then I changed back to the QR code URL that goes with test site 3
* remote command went through to Loop phone (but LoopFollow phone got a network error and stayed on carb screen) (add to summary)

marionbarker:
Design comments (my opinion only, for consideration)

On the main `Remote Settings` screen, there is a debug / info box at the bottom:
* having the `Current TOTP Code` on the main `Remote Settings` screen seems appropriate
* what about moving the rest of that to the `Configure Loop APNS Settings` screen instead

On the main `Remote Settings` screen, the `Setup Complete` turns green when all the rows on the `Configure Loop APNS Settings` screen are filled in - whether they are correct or not
* I deliberately modified the `APNS KEY` and the `Setup Complete` remains green
* When I try to send a remote with the "bad" key, I get the new error: `Failed generating JWT token, check APNS Key ID, APNS Key and Team ID`
* Can `Setup Complete` be tested before being turned green? And if that error occurs, turn it to incomplete again?

The APNS settings are not `Loop` specific. 
* Perhaps remove `LOOP ` from the rows in the `Configure Loop APNS Settings` screen for APNS and Developer. 
* Perhaps rename the screen to `APNS and Loop Settings` - because the QR image must be obtained from the Loop phone.

marionbarker:
## Summary

One intermittent issue:
* After I switched to a TestFlight Loop build for the phone feeding the Loop URL
   * The Production enabled setting on the LoopFollow phone kept changing to disabled and I'd get a network error, I'd enable production again and then request would go through.
   * I don't know the exact steps to make this happen - but it happened 3 times
   * Possibly something from one of my other test phones, but I did disconnect all other test phones from servicing that URL

One UI issue:
* The Remote Settings header row is transparent and sticky, in the current implementation, I think it should be opaque
* See graphic at the end of this comment

Overall, this worked as expected but I have a few observations:

**Loop**
* Should the label in Setting: Remote Settings be "Loop Remote Control" to be symmetric with "Trio Remote Control"
* There is no option to enter the time for the meal for Loop (I know this is supported because LoopCaregiver provides that option)
* Can the active override name be displayed on the Loop Remote Override screen? Similar to the way TRC works
* The done button is still entering a return instead of completing after pasting in the APNS Key

**Things that might be just the way they are:**
* I started with a QR code URL that did not match the test phone feeding the NS URL and LF was happy to pretend to send carbs to that phone (but nothing went through until the QR code was updated) - the OTP is obtained from the QR code URL (NS showed the error but LF did not)
* There is no "name" option for Loop - there used to be with the old NS method - not sure it's supported when going straight to Loop

**Trio**
* Not related to this PR, just summarize observations for later consideration
   * Trio remote meal
         * NS says command rejected because newer entries were logged (note shows up in LoopFollow)
         * Trio accepted the 5 g entry and the 5 g entry appears in LoopFollow
   * NS Label shows: xx g Remote meal command
      * Should this instead be remote name?
      * where `name` is entered by the user in the TRC Settings screen


## Test

Configuration:
* commit `66bf4e4`
* iPhone SE, 3rd gen running iOS 18.5
* this phone already has the APNS and TeamID keys from previous testing of this branch

Tests successful (see below for details)
* success: still works with Trio Remote Control using 0.5 Trio URL
* success: works with Loop Remote Control using Loop URL
* success: able to delete and add back credentials
   * Loop and Trio credentials that can be common, are common (TeamID, APNS KEY ID, APNS KEY) 

### Test Timeline

The notes below are in the order in which the tests were performed.

### Test Trio Remote Control

Make sure these modifications have not affected Trio Remote Control (TRC).
Some notations in the timeline below are not related to this PR. Just capturing them for later analysis.

* switch to a Trio 0.5.x URL
* check remote control settings: select Trio Remote Control
* add the shared secret
* enable an override from Trio phone
* Test TRC features
   * Overrides: current override is shown as active, cancel it
      * Missing required fields: check your remote settings: user
      * Added a name in the "user" row for TRC
      * Success: now able to cancel the override
      * Success: able to set a new override
   * Temp Target: able to send Temp Target
      * I did not use a named temp target (entered 110 with 30 min duration)
      * It shows up on Trio phone on the main screen but not in the adjustments, temp target screen of trio 
   * Success: Meal (configured for just carbs) - entries appear on Trio phone and NS
      * send 10 g for in the future
      * send 15 g for in the past
      * send 5 g for now
         * NS says command rejected because newer entries were logged (note shows up in LoopFollow)
         * Trio accepted the 5 g entry and the 5 g entry appears in LoopFollow
      * NS Label shows: xx g Remote meal command (Should this instead be remote name?)
   * Success: Meal (configured for bolus, fat/protein) - entries appear on Trio phone and NS/LoopFollow
   * Success: bolus  - entries appear on Trio phone and NS/LoopFollow

### Test Loop Remote Control

Modify the URL to Loop.
* that URL is not currently being serviced so I get a Not Looping notice
* Configure for Loop remote control
   * Setting: Remote Settings: automatically changed to None
   * Choose Loop (should this be Loop Remote Control for symmetry with Trio Remote Control?)
   * All my settings are retained from earlier tests
   * Note - did NOT update the QR code - this does not affect overrides
* Modify one of my test phones to service the Loop URL
   * add NS service, CGM and use rPi DASH pump
   * wait for NS to update and be reflected in LoopFollow
* add some named presets (overrides) to Loop phone and select one (sick), wait for it to appear in NS and LoopFollow
* Success: test sending an override
   * The choices for overrides found on test phone are seen in LoopFollow
   * selected biking and it appeared on Loop phone with sick cancelled as expected
   * shows up in NS promptly
   * shows up in information table of LF
   * does not show up as an active override on the override screen of LF, but there is a cancel active override button but maybe this is the way the interface is designed
   * (the TRC interface indicates which override is active)
* failed until I updated QR code, then succeeded: meal 
   * send 10 g carbs, duration 4 hr, note test
   * does not go through
   * update the QR code and now it goes through
   * got notification on Loop phone and can see carbs in history
   * there is not option to modify the time for the carb entry
* success: bolus, sent 1 U and it succeeded
   * got notification on Loop phone, can see bolus progress on Loop phone

### Reset credentials

While in Loop Remote Control Settings, delete all the credentials
* the rows for credentials are not allowing me to select all, having to delete a character at a time.
* I did figure out how to get one of them to let me select all, but it should be easier
* When I tapped on Remote for Loop, I was told the APNS was not configured

Switched to Trio URL
* all credentials, except shared secret, have been reset
* this is as expected

Switch back to Loop URL
* ensure all credentials can be added
* Note that developer TEAM ID got refilled
* The done button is still entering a return instead of completing after pasting in the APNS Key
* All other entries were nominal
* Sent remote bolus successfully

Install Loop onto test phone from TestFlight (before was built from Xcode)
* Do not modify the Production to be enabled on LoopFollow and attempt to bolus
* Once again - LF claims the command was sent but it did not go through
   * Note that NS had not yet been updated, later testing shows a network error when NS and LF don't agree on Production enable or not
* Modify Production to enabled and get a network error
* enable an override on loop phone (does not fix anything)
* modify carb ratio on loop phone (this fixes it)
* now I got a success - needed to force an update to NS after installing TF version
* but later I got a network error when requesting an override and found Production had been disabled again
* I was not able to reproduce this reliably but it happened more than once.

The header row for the Remote Settings screen probably needs work.
* It is transparent and sticky
* I assume this is so you can tap on Settings at any point
* It gets really busy when scrolling further down into the credentials section

loopfollow-pr434-ui

bjorkert:
[This](https://github.com/loopandlearn/LoopFollow/pull/434/commits/73f93994323bb4843faed47b67fe72d0328aef78) resolves the UI issue where the RemoteSettingsView had a transparent, sticky header.

The root cause was a nested NavigationView which created a separate, conflicting navigation context. I have refactored the view to fix this by:

Removing the unnecessary nested NavigationView.

Applying the view modifiers (like .navigationTitle, .onAppear, etc.) directly to the Form.

Setting the .navigationBarTitleDisplayMode(.inline) to ensure the title appears correctly within the main navigation bar immediately upon display.

This change aligns the view with standard SwiftUI practices, making it a proper subview of the main NavigationStack and completely fixing the header bug.

See picture: 
image

bjorkert:
>NS Label shows: xx g Remote meal command
>Should this instead be remote name?
>where name is entered by the user in the TRC Settings screen

It does not look very good in Nightscout, in trio history it's pretty nice. Including the user name could be an option.

codebymini:
> Can the active override name be displayed on the Loop Remote Override screen? Similar to the way TRC works

[36f4b25](https://github.com/loopandlearn/LoopFollow/pull/434/commits/36f4b25d62405f8aab84cfbf07c4fd4825f7a352)
Adds the current override to the top of the override screen and places the cancel override together with it

codebymini:
> There is no option to enter the time for the meal for Loop (I know this is supported because LoopCaregiver provides that option)
[392c4c5](https://github.com/loopandlearn/LoopFollow/pull/434/commits/392c4c534702ff3ae6a3a1231191dd3312f8aef4) Adds timepicker for future carbs and fix absorption time with default emojis from Loop

bjorkert:
>The done button is still entering a return instead of completing after pasting in the APNS Key

https://github.com/loopandlearn/LoopFollow/pull/434/commits/525950fa2f3f5f3ea93806ac8921ac26267b3561 fixes the issue where the Done button was inserting a newline instead of dismissing the keyboard in multiline fields (APNS Key field). The solution adds a keyboard toolbar with a Done button specifically for multiline TextEditor fields, while preserving the default Done behavior for single-line fields. The Done button now properly dismisses the keyboard without adding extra lines to the text.

bjorkert:
I think all issues should be fixed now, I think we can merge to dev if you are happy with the changes @marionbarker.

marionbarker:
## Summary

* The issues reported yesterday have been fixed with exception of better reporting when a remote command is marked as OK and then later fails
* Confirmed that LoopFollow log has the record for the remote commands issued
   * The logs may be a little hard to read, but probably best to be verbose for now (attached for reference)
   * Configuration: LoopFollow, Settings, Advanced Settings, Debug log is disabled 
* Any issues observed with Trio Remote Control in previous comment will be written up separately (not related to this PR).

## Test

Configuration uses most recent commit, `525950f`.
Repeat a few of the tests from yesterday where there was a request for more information.

Begin with Loop phone using a TestFlight build.
* Confirm that (with old code), the Production variable had returned to disabled
   * With the new code that fixes this issue, set variable to enabled
   * Make sure it stays enabled during the rest of the tests
   * Given the length of this test and the fact that it remained enabled, this has been fixed

Change to Override behavior
* activate an override (in this case do it remotely - this works)
* wait for LoopFollow to display the override in the plot
* tap on Remote, Override and observe the active override is displayed at the top of the screen
* Thanks for this update

UI update
* Observe that the header row for the Remote Settings screen is sticky and opaque
* It is much easier to see the `< Settings` button at upper left.
* Thanks for this update

Send Carbs with different time for consumption:
* Use icons to change absorption time
* Issue carbs in the past (about 20 min earlier)
   * success - arrived and were posted at the correct time 
* Issue carbs in the past (14 hours earlier)
   * LoopFollow - says must be within prior 12 hours
   * Change time to 12 hours
   * Success - arrived at Loop phone
* Issue carbs in the future
   * LoopFollow - says limit is 1 hour in the future
   * Change time to 1 hour
   * Success - arrived at Loop phone


Use the wrong OTP generator
* Modify the QR URL to one on a different phone than the one being controlled by the LoopFollow phone
* As expected, the LF Remote: Cancel Override works fine (this does not require a OTP)
* Repeated the issue from yesterday when sending a bolus
   * The OTP is incorrect (because I scanned a different QR code)
   * LoopFollow phone claimed success
   * Bolus was not sent to Loop phone
   * Error message as seen in NS is reported below
   * There is a grey dot on LoopFollow (with no bolus shown on LoopFollow) so maybe this is the best that can be expected
   * The error message - when viewed on LF phone is not readable - see graphic below

> Error: Password is incorrect: 700912. {"aps":{"alert":"Remote Bolus Entry: 0.05 U","interruption-level":"time-sensitive","content-available":1},"entered-by":"LoopFollow","otp":"700912","bolus-entry":0.050000000000000003,"remote-address":"LoopFollow","sent-at":"2025-08-05T13:08:47.306Z","notes":"Sent via LoopFollow APNS","expiration":"2025-08-05T13:13:47.306Z"}

This is what is seen on the LoopFollow phone after failed bolus when OTP is incorrect:

loopfollow-otp-wrong

## Log file

Here is the log file from today's test.

* [LoopFollow 2025-08-05.log](https://github.com/user-attachments/files/21600985/LoopFollow.2025-08-05.log)

codebymini:
> We have the same disappearing carbs problem as was reported with LoopCaregiver and Loop. The solution there was for LoopCaregiver to send a timestamp equal to the requested hh:mm with the current second included. This was done in this commit:
> 
> * [LoopKit/LoopCaregiver@ce5bde8](https://github.com/LoopKit/LoopCaregiver/commit/ce5bde8cabfd53a690fbc811dd25ef7af3148f96)
> 
> If carbs are uploading to Nightscout with the same hh:mm:ss, only one is kept. In other words, the multiple carb entries at the same time disappear in Nightscout but are there in Loop and can lead to confusion.
> 
> The carb entries in Nightscout vanish if they are at the same second. So if you set a time for entry of hh:mm:00 you will run into this. It is not uncommon for parents to say: carbs of xx at "lunch" time to get a prebolus going, then later maybe add more carbs at the same time.
> 
> Here's some information about this:
> 
> When using NS Careportal with Loop:
> 
> * https://loopkit.github.io/loopdocs/nightscout/remote-commands/#use-unique-times-for-remote-carbohydrate-entries
> 
> The LoopCaregiver code was modified to minimize the odds of this happening - by adding a second to the field (using current time on the phone). Although there is still a one in 60 chance of a duplicate hh:mm:ss for a remote carb. They is also true when using the Loop app directly.
> 
> In zulipchat:
> 
> * [#Loop Caregiver App > Remote Carbs Disappearing in Nightscout](https://loop.zulipchat.com/#narrow/channel/358458-Loop-Caregiver-App/topic/Remote.20Carbs.20Disappearing.20in.20Nightscout)

[c9fcf2e](https://github.com/loopandlearn/LoopFollow/pull/434/commits/c9fcf2ee54bb229dc29ca86d44932e7e59fc3c17) Implements the same logic for carbs as in Loop Caregiver

marionbarker:
## Test

I tested sending remote carbs to a Loop URL with the same hh:mm requested time from LoopFollow. The times were sent with different seconds so that they both appeared in Nightscout properly.

Thanks for the update.

marionbarker:
test build is ok.
LGTM

marionbarker:
One thing I noticed is the remote bolus minimum is set to 0.05 U.
And only 2 decimal places are allowed.
* Some Medtronic pumps allow 0.025 U increments when total amount is < 1 U.

Whatever bolus value is sent from LoopFollow LRC is modified to an allowed pump bolus value.

For pod:
* sent 2.54, delivered 2.50
* sent 0.12, delivered 0.10

For simulated pump (x23)
* sent 0.08 U, delivered 0.08 U
* can't enter the 0.025 U increments that are available to this pump

marionbarker:
Please add a check for Looping status. (I should have added this with yesterday’s comment.  I was too focused on documentation.)

I did not test Trio, but for Loop URL remote bolus offered an insulin amount that was ancient. 

Not sure best way to handle it. IMO, if currently Not Looping - any tap on a remote option should have a warning instead of showing the screen for that option. 

Yesterday I was offered an  ancient insulin value (from NS not Loop) for remote bolus when Not Looping.  That’s where the weird 2.54 U value example came from. When looping the rec bolus amount from Loop was always offered.

codebymini:
> Please add a check for Looping status. (I should have added this with yesterday’s comment. I was too focused on documentation.)
> 
> I did not test Trio, but for Loop URL remote bolus offered an insulin amount that was ancient.
> 
> Not sure best way to handle it. IMO, if currently Not Looping - any tap on a remote option should have a warning instead of showing the screen for that option.
> 
> Yesterday I was offered an ancient insulin value (from NS not Loop) for remote bolus when Not Looping. That’s where the weird 2.54 U value example came from. When looping the rec bolus amount from Loop was always offered.

[601ea95](https://github.com/loopandlearn/LoopFollow/pull/434/commits/601ea95c1cafe4070f78c65c8e93dd28b4fa05d7) Adds better warnings for bolus recommendation and also invalidates the recommendation if its older than 12 minutes

PR #433, opened 2025-07-08
feat: add user-configurable tab customization for tabs 2 and 4

bjorkert:
## Description

This PR implements user-configurable tab customization, allowing users to select what appears in tabs 2 and 4 of the tab bar. Previously, tab 2 was dynamically set based on whether remote was enabled, but now users have full control over their tab arrangement.

## Changes

- Added `TabSelection` enum to centralize tab configuration (display name, SF symbol, storyboard identifier)
- Added storage values for `tab2Selection` and `tab4Selection` with defaults (Alarms and Nightscout respectively)
- Created `TabCustomizationSettingsView` for users to configure their tab preferences
- Added migration step 2 to preserve existing user preferences (Remote users get Remote in tab 2, others get Alarms)
- Updated `MainViewController` to use the new dynamic tab system
- Fixed Nightscout tab enable/disable logic to work regardless of tab position
- Added missing storyboard identifier for NightscoutViewController

## User Experience

Users can now go to Settings → App Settings → Tab Customization to choose what appears in tabs 2 and 4 from:
- Alarms
- Remote  
- Nightscout

The fixed tabs remain:
- Tab 1: Home (fixed)
- Tab 3: Snoozer (fixed)
- Tab 5: Settings (fixed)

## Migration

Existing users will have their preferences preserved:
- If Remote was enabled: Tab 2 = Remote
- If Remote was disabled: Tab 2 = Alarms
- Tab 4 defaults to Nightscout for all users

## Testing

- [x] Verified tab customization works correctly
- [x] Tested migration preserves existing user preferences
- [x] Confirmed Nightscout tab is properly disabled when URL is empty
- [x] Verified all tab combinations work as expected

Closes #333

PR #437, opened 2025-07-17
More menu

### LINKED ISSUE #333: Feature Request: Select buttons in toolbar

bjorkert:
### feat: Implement Fully Dynamic Tab Bar with "More" Menu

**Closes #333**

This pull request introduces a fully dynamic tab bar, resolving the core feedback from the previous attempt (#433). The previous design could make features inaccessible if they weren't assigned to a tab. This new implementation ensures all wanted functionality remains reachable through a "More" menu.

#### Key Changes

* **Dynamic "More" Menu**: The last tab now intelligently switches between "Settings" and "More". If any feature (Alarms, Remote, Nightscout) is assigned to the "More" position, a `MoreMenuViewController` appears, listing those features and a link to Settings.
* **Flexible Tab Positions**: Users can now assign features to `Tab 2`, `Tab 4`, the `More Menu`, or hide them completely via a new `TabCustomizationSettingsView`. The UI cleverly prevents assigning multiple features to the same tab slot.
* **Robust Tab Controller**: `MainViewController` has been refactored to dynamically build the entire tab bar based on user settings. It gracefully handles UI updates, even when the "More" tab is removed while active.
* **Seamless Migration**: A new migration step automatically converts existing user setups to the new position-based system, preserving their established layout.

#### User Experience & Testing

1.  Navigate to `Settings > Tab Customization`.
2.  Assign "Alarms", "Remote", or "Nightscout" to different positions.
3.  Verify that if one or more items are set to "More Menu", the last tab becomes a "More" button.
4.  Verify that if no items are in the "More Menu", the last tab is the standard "Settings" button.
5.  Confirm that all features are accessible either directly from a tab or through the "More" menu, addressing the main concern of the previous PR.

This approach provides maximum flexibility without sacrificing access to features.
image
image
image


marionbarker (changes requested):
I take away the approval. I did more testing. The Done button is still only there sometimes.

Pretty sure the Done button on Tab Customization screen is missing following a setting where something is hidden.

* Settings, Tab Customization, Hide an item and tap Done
* Settings, Tab Customization - the actions can be taken and settings change but there is no Done button
   * Modify so the More menu is used and nothing is hidden, have to tap on one of the taskbar icons or tap on the Settings icons in upper left
   * Next time you go to Settings, Tab Customization, the Done button has returned

marionbarker:
## Tested with latest commit, d9269da

This now provides a cancel button upper left and apply button upper right for the Tab Settings screen. These buttons are rock solid in performance now.  Thanks.

marionbarker (approved):
Tested the last commit, c680029: Fix for snoozer tab position

Configuration, all adjustable tabs were hidden, so tab bar showed Home, Snoozer and Settings.

* Before the commit, when only 3 tabs were shown, the Settings screen opened when an alert was issued.

* After the commit, the Snoozer screen opened when an alert was issued.

LoopFollow/Settings/TabCustomizationModal.swift
```diff
@@ -0,0 +1,168 @@
...
+                        otherPositions: [alarmsPosition, nightscoutPosition]
+                    )
+                    .onChange(of: remotePosition) { _ in checkForChanges() }
+
+                    TabPositionRow(
+                        title: "Nightscout",
+                        icon: "safari",
+                        position: $nightscoutPosition,
+                        otherPositions: [alarmsPosition, remotePosition]
+                    )
```
codebymini, line 65: A reset to default button could be nice to have

codebymini (commented):
👍

PR #438, opened 2025-07-17
Ensure overrides are visible when predictions are disabled

### LINKED ISSUE #435: Remote overide command Issue with the disabled prediction indicator
When the prediction OREF algorithm is configured to a 0 time prediction, the remote command overrides do not update, and there is no option to eliminate the override from the remote command.

bjorkert:
Closes #435

This PR resolves an issue where setting the prediction time to 0 would cause overrides to be calculated with an end date equal to the current time. This made them look like they are ended on the graph and impossible to cancel with a remote command, as they didn't appear to be active.

The fix introduces a minimum display duration for overrides. The code now calculates the `maxEndDate` by using the greater of the user's `predictionToLoad` setting or a new `minimumFutureDisplayHours` constant (defaulting to 1 hour).

This ensures that a long override is always rendered for at least one hour into the future, making it visible and actionable even when prediction lines are turned off.

### DIFF
--- a/LoopFollow/Controllers/Nightscout/Treatments/Overrides.swift
+++ b/LoopFollow/Controllers/Nightscout/Treatments/Overrides.swift
@@ -21,7 +21,10 @@ extension MainViewController {
         }
 
         let now = Date().timeIntervalSince1970
-        let maxEndDate = now + Storage.shared.predictionToLoad.value * 3600
+        let minimumFutureDisplayHours = 1.0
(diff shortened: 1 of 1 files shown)


marionbarker:
Test again. This is definitely better. Graphic below shows `minimumFutureDisplayHours` at 1 hour on the left and 0.25 hour on the right. In both cases, the Hours of Prediction is set to 0.25 hours.

lf-pr438

Issue #440, opened 2025-07-19
Bigger bg text in snoozer view and show display name in snoozer view


bjorkert:
The Snoozer view needs two key improvements to enhance readability and user context:

Increase BG Text Size: The primary blood glucose value (bgText) is difficult to read from a distance. Its font size and frame height should be increased to make it the most prominent element on the screen.

Show Display Name: Users with multiple setups need to quickly identify which one they are viewing. An option to show the custom app display name (from CFBundleDisplayName) should be implemented. This name should be visible but secondary to the main BG data.

PR #441, opened 2025-07-19
Enlarge BG text and add optional display name

bjorkert:
This PR introduces several enhancements to the Snoozer view to improve at-a-glance readability and provide better user context, addressing Issue #440 .

**Key Changes:**

* **Larger BG Text:** The font size and frame for the BG value have been increased, making it the primary focal point of the view.
* **Optional Display Name:** The app's display name can now be shown in the `rightColumn`, which works well in both portrait and landscape modes. This is controlled by the `showDisplayName` setting.
* **Adaptive Landscape Layout:** In landscape mode, the trend arrow and delta value are now displayed side-by-side for a more compact and readable layout.

### DIFF
--- a/LoopFollow/Snoozer/SnoozerView.swift
+++ b/LoopFollow/Snoozer/SnoozerView.swift
@@ -29,12 +29,12 @@ struct SnoozerView: View {
                     if isLandscape {
                         HStack(spacing: 0) {
                             leftColumn(isLandscape: true)
-                            rightColumn
+                            rightColumn(isLandscape: true)
                         }
                     } else {
                         VStack(spacing: 0) {
                             leftColumn(isLandscape: false)
-                            rightColumn
+                            rightColumn(isLandscape: false)
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


bjorkert:
65F64591-15BB-4A8E-8A5F-6F6F5F13153D_4_5005_c
Simulator 2025-07-19 18 45 06
Simulator 2025-07-19 18 44 32

marionbarker:
## Summary

Overall - works as coded.
Questions
- seems like the name on snoozer should be in bigger font too (I'd make the name bigger than the emoji) and in portrait maybe above the glucose
-  I think the default for a new instance (not a rebuild) of LoopFollow should configure the name to be enabled

## Test

Looks good. These results were with LoopFollow, General Settings, Show Display Name enabled. After I captured the screenshots, I enabled Snoozer emoji too - I like that. (I know you added it a while ago.) Snoozer screen still good with emoji and name both enabled.

Portrait with dev (left) and with PR (right):

lf-441-portrait

Landscape with dev (upper) and with PR (lower):

lf-441-landscape

marionbarker (approved):
I reviewed the code and tested. I remarked on some possible modifications. But I approve as is if you decide not to include these suggestions.

marionbarker (approved):

PR #442, opened 2025-07-20
Larger and more visible Snooze Button

mikebash:
In case this is of interest to the broader community, in addition to increasing the size of the BG in the Snoozer screen (which is being addressed in a different pull request), I also increased the size of the snooze button and adjusted the color to stand out at night.  Fully recognize that a large white button could be considered overly bright for others but I find it helpful for when I do not have my glasses on in the dark.

Team - feel free to reject and/or alter any of this.  Simply sharing in case it's helpful.

### DIFF
--- a/LoopFollow/Snoozer/SnoozerView.swift
+++ b/LoopFollow/Snoozer/SnoozerView.swift
@@ -47,7 +47,7 @@ struct SnoozerView: View {
     private var leftColumn: some View {
         VStack(spacing: 0) {
             Text(bgText.value)
-                .font(.system(size: 220, weight: .black))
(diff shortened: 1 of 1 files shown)


bjorkert:
Hey @mikebash! 

Thanks for catching this accessibility issue - the struggle without glasses at night is real! Your bigger button idea is perfect and I've incorporated it into #443 with an orange color (like Apple's snooze button) for better contrast.

Really appreciate you spotting this and taking the time to PR it. Please keep these practical improvements coming - this is exactly the kind of real-world feedback that makes the app better!

Thanks! 🙏

PR #445, opened 2025-08-07
Enable APNS feedback for remote commands

bjorkert:
### Description

This pull request implements the client-side functionality for receiving real-time status notifications for remote commands sent to Trio or Loop. This complements the backend changes and completes the feature, allowing users to get immediate push notification feedback confirming the success or failure of their actions.

To achieve this, the app is now capable of:
1.  Registering for and receiving push notifications from Apple's Push Notification Service (APNS).
2.  Constructing a `return_notification` payload to attach to every outgoing remote command.
3.  Handling scenarios where Loop Follow and the target app (Trio/Loop) are built with different Apple Developer Team IDs.

#### 1. Push Notification Registration and Handling
* **`AppDelegate`**: The app now registers for remote notifications on launch. New delegate methods have been implemented to:
    * Securely capture and store the device token from APNS upon successful registration.
    * Log any registration failures.
    * Receive and process incoming feedback notifications from Trio, displaying them as alerts to the user even when the app is in the foreground.
* **Project Configuration**: The project's `.entitlements` and `Info.plist` files have been updated with the required `aps-environment` and `remote-notification` background mode capabilities, which are necessary for receiving push notifications.

#### 2. Dynamic Return Payload Creation
* A new method, `createReturnNotificationInfo()`, has been added to the `LoopAPNSService` and `PushNotificationManager`. This method is now called before any remote command is sent.
* It constructs a `return_notification` object containing all the information the remote peer (Trio) needs to send a notification back to *this specific* Loop Follow app instance. This payload includes:
    * The Loop Follow app's unique device token.
    * The build environment (production/development).
    * The necessary APNS credentials (Team ID, Key ID, and APNS Key).

#### 3. Support for Mismatched Developer Teams
A core part of this implementation is handling cases where Loop Follow is built with a different Apple Developer account than the target Loop/Trio app.
* **Build Script**: The build script now captures the `DEVELOPMENT_TEAM` ID at compile time and stores it in `BuildDetails.swift`.
* **Conditional Logic**: The app compares its own Team ID against the Team ID of the target remote app.
* **New Configuration UI**: If the Team IDs are different, a new "Return Notification Settings" section dynamically appears in the `Remote Settings` view. This prompts the user to enter the APNS Key and Key ID associated with their Loop Follow developer account. These credentials are then used exclusively for the return notification path, ensuring that Trio can authenticate correctly to send the feedback notification. If the Team IDs are the same, the primary APNS credentials are reused for the return path, and no extra configuration is needed.

### DIFF
--- a/LoopFollow/Application/AppDelegate.swift
+++ b/LoopFollow/Application/AppDelegate.swift
@@ -40,10 +40,62 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         _ = BLEManager.shared
 
+        // Register for remote notifications
+        DispatchQueue.main.async {
+            UIApplication.shared.registerForRemoteNotifications()
+        }
+
         return true
     }
 
     func applicationWillTerminate(_: UIApplication) {}
+
+    // MARK: - Remote Notifications
+
+    // Called when successfully registered for remote notifications
+    func application(_: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
+        let tokenString = deviceToken.map { String(format: "%02.2hhx", $0) }.joined()
+
+        Observable.shared.loopFollowDeviceToken.value = tokenString
+
+        LogManager.shared.log(category: .general, message: "Successfully registered for remote notifications with token: \(tokenString)")
+    }
+
+    // Called when failed to register for remote notifications
+    func application(_: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
+        LogManager.shared.log(category: .general, message: "Failed to register for remote notifications: \(error.localizedDescription)")
+    }
+
+    // Called when a remote notification is received
+    func application(_: UIApplication, didReceiveRemoteNotification userInfo: [AnyHashable: Any], fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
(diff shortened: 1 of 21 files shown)


bjorkert:
I've pushed the latest updates to this PR, which now implements end-to-end encryption flow for remote commands.

Instead of sending the shared secret for verification, I now use it as a key to encrypt the entire command payload. Here are the technical details of the implementation:

Encryption Algorithm: I've implemented this using AES-256-GCM. The GCM (Galois/Counter Mode) is important because it's an authenticated encryption mode, meaning it not only ensures confidentiality but also provides data integrity, protecting against tampering.

Key Derivation: The encryption key is derived from the existing shared secret (the GUID). I pass the secret through a SHA-256 hash to produce the consistent 256-bit key required for the AES algorithm.

Nonce Handling: For each message sent, a new, cryptographically secure 12-byte nonce (Initialization Vector) is generated. This ensures that even identical commands result in unique ciphertext.

Authentication: The shared secret is no longer sent over the network. Authentication is now implicit: if Trio can successfully decrypt the payload, the sender is proven to be authentic.

Important: This is a breaking change. Current versions of LoopFollow and Trio dev will not be compatible with these new versions. Both apps must be updated together for the remote control feature to work.

bjorkert:
Important step for Browser Builders: Now that LF receives push notifications, you need to enable the Push Notifications capability at the Apple Developer site.

PR #449, opened 2025-08-11
Fix(alarm): Correct logic for missed reading alert

bjorkert:
This PR fixes a bug where the Missed Reading alert would not trigger when CGM data became stale.

### Problem

The alert was being blocked by a global check that required all `isBGBased` alarms to have *recent* BG data. This directly conflicted with the purpose of the Missed Reading alert.

### Solution

-   The alarm remains classified as `isBGBased` to correctly use the system's debounce logic (`lastBGAlarmTime`).
-   A targeted exception, `alarm.type != .missedReading`, is added to the "recent data" check, allowing the alarm to fire as intended.
-   The condition logic is also updated to use the correct `now` timestamp for consistency.

### DIFF
--- a/LoopFollow/Alarm/AlarmCondition/MissedReadingCondition.swift
+++ b/LoopFollow/Alarm/AlarmCondition/MissedReadingCondition.swift
@@ -9,7 +9,7 @@ struct MissedReadingCondition: AlarmCondition {
     static let type: AlarmType = .missedReading
     init() {}
 
-    func evaluate(alarm: Alarm, data: AlarmData, now _: Date) -> Bool {
+    func evaluate(alarm: Alarm, data: AlarmData, now : Date) -> Bool {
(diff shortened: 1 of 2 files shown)


marionbarker (approved):
## code review

Approve from code review.

## Test

Build LoopFollow dev on test phone
* configure to follow Loop test phone running closed loop with rPi DASH simulator and CGM readings from Nightscout (fed by upload from Mac), make sure Loop is green and LoopFollow is following
* interrupt glucose readings sent to Nightscout
* set alarm to alert for missed readings after 11 minutes
* wait 15 minutes for alarm (to not happen)
   * last glucose reading in NS at 10:48
   * get distracted - it is now 11:13 and there is no Missed Reading Alert
   * this demonstrates the error that is being fixed

Build with this version of code
* get missed reading alarm almost immediately

PR #466, opened 2025-10-04
Recommended bolus for Trio

bjorkert:
# Trio Remote Control: Recommended Bolus on Bolus Screen

This PR adds a flow for using the **Recommended Bolus** in the Trio Remote Control bolus view.

### What’s included
- Shows a **Recommended Bolus** chip when a fresh recommendation is available.
- Hides the chip if the value is negative/zero/tiny (<0.05U) or older than ~12 minutes.
- Color cue for freshness:
  - Yellow when recently calculated.
  - Red warning if ~5–12 minutes old.
- Tapping the chip:
  - Applies directly when fresh.
  - Prompts a caution dialog if older than ~5 minutes.

### Why
Make it quick to use a safe, recent recommendation—and avoid nudging users toward stale or invalid values.

### DIFF
--- a/LoopFollow/Controllers/Nightscout/DeviceStatusOpenAPS.swift
+++ b/LoopFollow/Controllers/Nightscout/DeviceStatusOpenAPS.swift
@@ -45,6 +45,7 @@ extension MainViewController {
                 updatedTime = parsedTime
                 let formattedTime = Localizer.formatTimestampToLocalString(parsedTime)
                 infoManager.updateInfoData(type: .updated, value: formattedTime)
+                Observable.shared.enactedOrSuggested.value = updatedTime
             }
 
             // ISF
--- a/LoopFollow/Remote/TRC/BolusView.swift
+++ b/LoopFollow/Remote/TRC/BolusView.swift
@@ -7,36 +7,45 @@ import SwiftUI
 
 struct BolusView: View {
     @Environment(\.presentationMode) private var presentationMode
+
     @State private var bolusAmount = HKQuantity(unit: .internationalUnit(), doubleValue: 0.0)
-    private let pushNotificationManager = PushNotificationManager()
     @ObservedObject private var maxBolus = Storage.shared.maxBolus
 
-    @FocusState private var bolusFieldIsFocused: Bool
+    @ObservedObject private var deviceRecBolus = Observable.shared.deviceRecBolus
+    @ObservedObject private var enactedOrSuggested = Observable.shared.enactedOrSuggested
 
+    @FocusState private var bolusFieldIsFocused: Bool
     @State private var showAlert = false
     @State private var alertType: AlertType? = nil
     @State private var alertMessage: String? = nil
     @State private var isLoading = false
     @State private var statusMessage: String? = nil
 
+    private let pushNotificationManager = PushNotificationManager()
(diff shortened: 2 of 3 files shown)


bjorkert:
LoopFollow currently does not know what pump is used, and therefore not what increments used. I plan to introduce a setting for it. I believe it can be fetched from Loop.

salguccio:
I wanted to thank you for making this extension available to all of us. I assure you, many parents use it. I hope it'll be added to the official versions of Trio and LoopFollow so that when an update is released by the developers, it can be downloaded without having to wait for "bjorkert" to make the appropriate changes.
THANK YOU...!!! THANK YOU FOR MAKING ALL PARENTS AVAILABLE TO HAVE THE BOLO TIP DIRECTLY IN THE FOLLOW APP... YOU'RE GREAT...!!! THANK YOU...

marionbarker:
## Test

This test was not successful.
* When bolus increment of 0.1 U is set, a requested bolus that included a 0.05 U increment was accepted
* LoopFollow displayed it rounded down but sent it to Trio as rounded up

### Configuration

Trio phone is an SE 2nd gen phone running iOS 18.7.2
* Trio version is 0.6.0.16, commit `3080f0429`
* Medtronic 523 pump is connected to the Trio phone
* Feature, Remote Control is enabled

LoopFollow phone is an iPhone 15 running iOS 18.7.2
* branch `recommended-bolus-trio`, commit 54445998
* configured LF for Trio Remote Control
* new line in the Debug/Info section
   * Bolus Increment 0.100 U 

### Remote Bolus

**The bolus increment is not being honored**
* A bolus of 0.05 U was accepted when entered in the LoopFollow remote control bolus field
* LF asked confirm bolus of 0.0 U - I confirmed
* The Trio phone, accepted a bolus of 0.1 U

Repeat the test with same result

Try a larger bolus so I have time to watch it be delivered on Trio and on pump itself
* A bolus of 0.85 U was accepted
* LF asked to confirm bolus of 0.8 U - I confirmed
* The Trio phone, accepted a bolus of 0.9 U
* A bolus of 0.725 was accepted - in this case LF rounded down and Trio bolused the rounded down amount of 0.7 U

### Recommended Bolus

* Delete the history of the 4 remote boluses from Trio phone so a bolus will be recommended
* Wait for next cycle, started temp basal of 2.4 U/hr and started a bolus - canceled it part way / resumed delivery
   * (With a Medtronic pump, canceling delivery means pump gets suspended and must manually resume)
* Rec bolus is now 2.8 U
* Tap on remote bolus and 2.8 U is recommended - but there is an old calculation warning, send it anyway
   * automatically reduced to 1 U because that's the max configured for this LF phone
   * cancel bolus on Trio phone once it started
* Modify LF configuration to enable a 5 U bolus to be sent
* Rec Bolus is 2.05
   * tap on remote bolus, suggests 2.0 U but **says it was calculated 10 minutes ago** - but I checked and the 2.05 U agrees with what the Trio phone is recommending.
   * The Trio phone (and the pump) showed a 2.0 U bolus starting - cancel it once it started
* Come back later and now the rec bolus does say calculated 1 min ago, so maybe it is updating properly: 
   * 2.2 U rec at LF screen
   * 2.25 U rec on the Trio screen

PR #467, opened 2025-10-04
BLE: Delay reconnect 2s after disconnect

bjorkert:
### Improve BLE reconnect behavior after disconnect

**What**
- On `didDisconnectPeripheral`, schedule reconnect after 2 seconds.
- Cancel any outstanding connect-timeout timer.
- Only reconnect if Central is powered on and the peripheral isn’t already `.connecting`/`.connected`.

**Why**
- Some Anubis transmitters briefly stay “alive” or do a quick connection flurry after a disconnect. An immediate reconnect can keep the radio active and increase battery use. A short 2s debounce lets the transmitter finish powering down/settling, reducing churn and power draw.


codebymini (approved):
Approved based on code-review

PR #469, opened 2025-10-05
Add small delay to TOTP warning when sending carbs/insulin

codebymini:
Enhance TOTP warning handling in LoopAPNS views by adding a delay before displaying the warning to prevent flickering. Updated both LoopAPNSBolusView and LoopAPNSCarbsView to include a new state variable for managing the warning display.


marionbarker (approved):
Having the TOTP blocked message pop up after a successful command and before returning to the Loop Remote Control screen has bothered me in the past.

I reviewed the code and testing this in combinations with PR 445, 468 and 469 (this one).

LGTM

PR #473, opened 2025-10-12
RemoteSettings: exclude Loop APNS from Team ID mismatch check

bjorkert:
## Summary
`areTeamIdsDifferent` should only compare Team IDs for **TRC**.  
This moves `.loopAPNS` out of the comparison path and into the “not applicable” path, preventing false “different Team IDs” states when configuring Loop APNS.


marionbarker (approved):
Approved by review and test build.
LGTM

PR #491, opened 2025-12-16
Remove icloud backups

codebymini:
Modify the import/export feature to remove the requirement for adding an iCloud container.
Using an iCloud Container adds complexity to the build process for those who use GitHub actions because the LoopFollow Identifier requires modifications that we don't know how to automate.

### DIFF
--- a/LoopFollow/Settings/DexcomSettingsView.swift
+++ b/LoopFollow/Settings/DexcomSettingsView.swift
@@ -47,7 +47,7 @@ struct DexcomSettingsView: View {
                 HStack {
                     Image(systemName: "square.and.arrow.down")
                         .foregroundColor(.blue)
-                    Text("Import Settings from QR Code or iCloud")
+                    Text("Import Settings from QR Code")
                 }
             }
         }
--- a/LoopFollow/Settings/ImportExport/ImportExportSettingsViewModel.swift
+++ b/LoopFollow/Settings/ImportExport/ImportExportSettingsViewModel.swift
@@ -36,12 +36,14 @@ class ImportExportSettingsViewModel: ObservableObject {
 
     enum ExportType: String, CaseIterable {
         case nightscout = "Nightscout Settings"
+        case dexcom = "Dexcom Share Settings"
         case remote = "Remote Settings"
         case alarms = "Alarm Settings"
 
         var icon: String {
             switch self {
             case .nightscout: return "network"
+            case .dexcom: return "network"
             case .remote: return "antenna.radiowaves.left.and.right"
             case .alarms: return "bell"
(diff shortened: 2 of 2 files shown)


marionbarker:
## Questions

1. If we are going to add this, can we figure out a way to automate adding the capability to the Identifier?
2. How many people will actually need this vs using the transfer to a new phone method?
3. What things cannot be transferred with the QR images that are transferred with the iCloud  Container?
4. Even though I did not configure my container before building, after the successful Browser Build - my Identifier showed the container was configured and I cannot figure out how to delete the existing container - so I'm not sure the success was "real"

## Test Browser Build

1. Delete iCloud Container capability from my Apple Dev account for LoopFollow.
5. Add this branch to the docs-test GitHub org.
6. Make this branch the default branch.
7. Run all 4 actions.
   * ✅ Validate Secrets
   * ✅ Add Identifiers
      * Examine capabilities: only push notifications is added as a capability
      * Revoke my Distribution Certificate 
   * ✅ Create Certtificates
      * Observe that nuke certs was triggered 
   * ❌ Build LoopFollow
      * Still need to have the iCloud capability
      * Add iCloud to Identifier but do not Configure it
   * ❓ Run Build only (confirms certificate updated with new capability)
      * See question 4 above 


Exact error message for build without adding to Identifier:

> :error file=/Users/runner/work/LoopFollow/LoopFollow/LoopFollow.xcodeproj::Provisioning profile "match AppStore com.***.LoopFollow" doesn't support the iCloud capability
> /Users/runner/work/LoopFollow/LoopFollow/LoopFollow.xcodeproj: error: Provisioning profile "match AppStore com.***.LoopFollow" doesn't include the com.apple.developer.ubiquity-kvstore-identifier entitlement. Profile qualification is using entitlement definitions that may be out of date. Connect to network to update. (in target 'LoopFollow' from project 'LoopFollow')

marionbarker (changes requested):
There are references to the cloud storage in the code I do not know how to test or access.

There is no way to export a QR code for Dexcom settings - I think that should be added.

The Nightscout Settings and Dexcom Settings screens each have a button that states:
`Import Settings from QR Code or Cloud` but both take the user to the Import/Export settings screen that only has import from QR code.

The Remote Screen has a button for Import/Export with no mention of Cloud.

marionbarker:
## Test

The labels for importing now refer to just the QR codes as expected.

### Test the QR Codes

✅ Nightscout Settings import and export both worked
✅ Dexcom Settings import and export both worked
✅ Remote Settings import and export both worked
✅ Alarms  import and export were tricky but move the camera to get the right focus and they do import

PR #494, opened 2026-01-02
Remove ISF unit detection

bjorkert:
This auto detection of ISF is not needed for Trio dev since the unit is always milligramsPerDeciliter.


marionbarker (approved):
Approved from code review and test.

## Test

* build this branch and view a Trio URL
* configure the Trio source so that ISF shows an enacted value
* modify the therapy settings on Trio phone to use ISF = 9 mg/dL/U
* confirm the ISF in the Information display shows the correct values of 9 mg/dL/U

PR #504, opened 2026-01-17
Add remote response notification for Loop

codebymini:
### Summary
This change makes the app properly accept APNS notifications sent back from a looping phone when a remote carbs and/or bolus request is issued. It ensures the notification path is valid, properly configured, and aligned with the remote-command settings flow so the response can be received reliably.
### Details
* Aligns the remote command setup so APNS credentials and environment are handled consistently when communicating with the looping phone.
* Ensures the app is eligible to receive the push response for remote carbs/bolus actions.
Adds supporting adjustments in the remote settings and notification handling flow to reduce failure cases in end‑to‑end remote responses.

### DIFF
--- a/LoopFollow/Application/AppDelegate.swift
+++ b/LoopFollow/Application/AppDelegate.swift
@@ -77,25 +77,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 let title = alert["title"] as? String ?? ""
                 let body = alert["body"] as? String ?? ""
                 LogManager.shared.log(category: .general, message: "Notification - Title: \(title), Body: \(body)")
-
-                // Check if this is a command response from Loop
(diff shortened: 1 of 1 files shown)


marionbarker:
## Test

Test both with just this branch, and with this branch updated with the latest version of dev (4.5.0) locally.

* build to a test phone which is examining my personal Loop URL
* I can send carbs and bolus and both go through immediately to my loop phone.

The LoopFollow plot reflects the remote entries once the data is uploaded from Loop to Nightscout and then downloaded from Nightscout to LoopFollow.

The comment for this PR indicates: `Ensures the app is eligible to receive the push response for remote carbs/bolus actions.`

Next step will be a modification on the Loop side for this to return a message via APNS to the LoopFollow phone.

PR #510, opened 2026-01-24
Add support for multiple bundle IDs

achkars:
This change helps create multiple versions of Loop Follow for distribution.

Steps:

- Create a new fork/branch
- In Config.xcconfig, modify the BUNDLE_IDENTIFIER variable.
- Build normally using GH actions


THERF-dotcomg3qg (approved):

bjorkert:
Closing this in favor of #690, which solves the same problem. Rather than adding a new BUNDLE_IDENTIFIER key, it ended up wiring the existing app_suffix mechanism (already used for the sister apps) into the Fastfile, so it stays consistent with how bundle IDs are handled elsewhere in the project. Appreciate you taking a crack at this.

PR #516, opened 2026-01-31
New stats

codebymini:
This PR adds a new Statistics experience, a full Treatments workflow, and a more flexible Tab/Menu customization system.

- Introduces a new stats screen with date-range support, core glucose/insulin metrics, and visualizations for AGP, TIR (including tight range toggle), and GRI.
- Adds data availability handling and stats data-fetching logic so calculations can run against the selected date range.
- Adds a Treatments screen with grouped history, pull-to-refresh, pagination (“Load more”), and treatment detail views with contextual Loop/device data when available.
- Adds treatment filtering so users can quickly narrow to carbs, basal, and bolus-related entries (including automatic/SMB/manual distinctions).
- Refactors tab customization to dynamic drag/reorder behavior, letting users control which items appear in the top tabs vs Menu.
- Updates storage and migration logic to support the new tab-position model and preserve compatibility for existing users.
- Updates navigation/menu integration so Stats and Treatments can be accessed cleanly via tabs/menu depending on user configuration.

### DIFF
--- a/LoopFollow/ViewControllers/MainViewController.swift
+++ b/LoopFollow/ViewControllers/MainViewController.swift
@@ -308,6 +308,20 @@ class MainViewController: UIViewController, UITableViewDataSource, ChartViewDele
             }
             .store(in: &cancellables)
 
+        Storage.shared.graphTimeZoneEnabled.$value
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.infoTable.reloadData()
+            }
+            .store(in: &cancellables)
+
+        Storage.shared.graphTimeZoneIdentifier.$value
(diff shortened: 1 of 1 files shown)


marionbarker:
For Loop, at least, stats should go by whole days. Otherwise Daily carbs and TDD will be off. This particular Nightscout URL is new with only a few days of history. 

Might not matter for most people as far as first day is concerned although I think it should be first full day in range and I think it should stop at midnight of previous day. Current time is 2026-02-08 at 8:30 am.

For reference 
* first data in the URL is 13:00 local on 2026-02-06
* my TDD is ~33 U

Possibly there should be a current day option that has values since midnight. 

IMG_8740

PR #518, opened 2026-02-12
Add error message when no device authentication has been set up

codebymini:
Improves device authentication error handling for remote bolus and meal actions by surfacing a clear, actionable message when authentication is unavailable (for example, no passcode/biometric setup).


bjorkert (approved):

marionbarker:
## Test

Built Loop main (4.6.0)

removed passcode from device
* tested that I could send remote carb, it appeared in loop and real-time response returned
* tested that when I attempt to send a bolus, nothing happens

build using the code in this PR
* tested that I could send remote carb, it appeared in loop but no real-time response so I think this branch is missing that feature
* tested that when I attempt to send a bolus, clear error message received

IMG_0260

PR #522, opened 2026-02-23
Add dark and tinted icons

codebymini:
- Adds dedicated iOS app icon variants for Any, Dark, and Tinted appearances (1024.png, 1024-dark.png, 1024-tinted.png).
 - Updates the AppIcon asset configuration to use the single-size iOS format.
 - Removes legacy per-size PNG icon files and updates convert_icon.sh to generate the new Any/Dark/Tinted outputs plus Contents.json.

Apple now supports generating iOS/iPadOS icon variations from a single 1024×1024 asset (and explicitly supports Single Size with Any/Dark/Tinted variants), so the old manually maintained small-size PNG set is no longer required:
[https://developer.apple.com/documentation/xcode/configuring-your-app-icon](https://developer.apple.com/documentation/xcode/configuring-your-app-icon)


MikePlante1 (approved):

MikePlante1:
I did get linting/fixems in 68 files when I built with Xcode 26.2, though.

PR #524, opened 2026-02-27
Align Fat/Protein order in TRC to Trio v0.6.0.51

MikePlante1:
Also display banner to notify users of the swap. Don't show again once dismissed and don't show for new users.

| Light | Dark |
|---|---|
|  |  |


bjorkert (approved):
Clean, well-thought-out PR. Migration logic correctly distinguishes existing vs new users, banner only shows when fat/protein is enabled, and TODO cleanup dates are clear. LGTM.

PR #530, opened 2026-02-28
Remove unnecessary iOS version checks

bjorkert:
## Summary
- Remove `if #available(iOS 16.0, *)` guard in `GeneralSettingsView.swift` — always true since the deployment target is iOS 16.6
- Remove `#if swift(>=5.9)` / `#else` wrapper in `EKEventStore+Extensions.swift` — Xcode versions supporting iOS 16.6 targets guarantee Swift 5.9+
- The `if #available(iOS 17, *)` check for `requestFullAccessToEvents` is kept since it guards an API above the deployment target

### DIFF
--- a/LoopFollow/Extensions/EKEventStore+Extensions.swift
+++ b/LoopFollow/Extensions/EKEventStore+Extensions.swift
@@ -4,26 +4,16 @@
 import EventKit
 import Foundation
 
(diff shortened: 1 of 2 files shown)


codebymini (approved):
LGTM!

PR #536, opened 2026-03-10
Fix NUKE_CERT variables to be case insensitive in bash

bjorkert:
## Summary
- Use `${VAR,,}` bash parameter expansion for case-insensitive comparisons instead of introducing `_LC` helper variables (supersedes #528)
- Fix bug where `FORCE_NUKE_CERTS` was compared as the literal string `vars.FORCE_NUKE_CERTS` instead of the actual variable value
- Add missing `FORCE_NUKE_CERTS=${{ vars.FORCE_NUKE_CERTS }}` assignment so bash has access to the variable
- No changes to the `nuke_certs` job `if:` condition since GitHub Actions `==` is already case-insensitive

## Test plan
- [ ] Set `ENABLE_NUKE_CERTS` to `True`, `TRUE`, and `true` — all should be recognized
- [ ] Set `FORCE_NUKE_CERTS` to `True`, `TRUE`, and `true` — all should trigger the force warning annotation
- [ ] Set `FORCE_NUKE_CERTS` to `false` or unset — no force warning annotation
- [ ] Verify the `nuke_certs` job still triggers correctly based on the GitHub Actions `if:` expression

### DIFF
--- a/.github/workflows/create_certs.yml
+++ b/.github/workflows/create_certs.yml
@@ -59,8 +59,9 @@ jobs:
         id: set_output
         run: |
           CERT_STATUS_FILE="${{ github.workspace }}/fastlane/new_certificate_needed.txt"
-          ENABLE_NUKE_CERTS=${{ vars.ENABLE_NUKE_CERTS }}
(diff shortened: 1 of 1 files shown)


marionbarker (approved):

PR #556, opened 2026-03-19
Tab migration

bjorkert:
Fix snoozer tab disappearing when upgrading from main to dev.

The `snoozerPosition` key never existed in UserDefaults on main (snoozer was hardcoded at position 3). On dev, the StorageValue default for `snoozerPosition` is `.menu`, so an unstored value silently becomes `.menu` and snoozer drops out of the tab bar.

Additionally, `nightscoutPosition` may not have been stored on main (step 2 set it to the same value as the default), causing it to pick up dev's different default (`.position3`) and collide with where snoozer needs to go.

**Changes:**
- Rewrite step 5 to use `snoozerPosition.exists` to detect users coming from main, and restore snoozer to position 3 after evicting any unstored occupant
- Add the same collision guard to step 3 for users arriving with `migrationStep < 3`
- Add `migratePositionExists(for:)` helper scoped to migration code

### DIFF
--- a/LoopFollow/Storage/Storage+Migrate.swift
+++ b/LoopFollow/Storage/Storage+Migrate.swift
@@ -5,17 +5,31 @@ import Foundation
 
 extension Storage {
     func migrateStep5() {
-        let isLegacyDefaultTabBarOrder = homePosition.value.normalized == .position1
-            && alarmsPosition.value.normalized == .position2
-            && remotePosition.value.normalized == .position3
-            && nightscoutPosition.value.normalized == .position4
-            && snoozerPosition.value.normalized == .menu
+        // Users upgrading from main had snoozer hardcoded at position 3, but that
+        // key never existed in UserDefaults. On dev the default for snoozerPosition
+        // is .menu, so an unstored value silently becomes .menu and snoozer
+        // disappears from the tab bar.
+        //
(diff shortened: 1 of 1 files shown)


marionbarker:
## Test

❌ This PR does not solve the migration problem.

SE 3rd gen running iOS 26.3.1

### Configuration

build LoopFollow v4.6.0 onto test phone.
* rearrange the toolbar (many times and can't reproduce, just list arrangement in next row until problem was reproduced)
   * set everything hidden; observe home, snooze, settings
   * this finally worked to reproduce the issue

build LoopFollow v4.6.14 onto test phone.
❓ Observe only 4 tabs show in the toolbar: Home, Snoozer, Treatments, Menu << not the desired behavior

build LoopFollow with the code in this PR onto test phone.
❌ Still only 4 tabs

build v4.6.0 again and then go straight to tab migration branch without going through v4.6.14
* v4.6.0: observe home, snooze, settings
* build tab-migration branch over main branch
* ❌ tab-migration still does not migrate properly

### Screenshot

The screenshot below was from:
1. build main
   * rearrange tabs so everything that can be hidden is hidden
   * clean build folder
2. build tab-migration, commit 89f7261
   * observe toolbar shows 4 tabs: home, snoozer, treatments, Menu
   * take screenshot of Menu, Settings, Tab screen


IMG_7546

PR #562, opened 2026-03-21
Clamp prediction values to 39-400 mg/dL

### LINKED ISSUE #495: Graph scale with high COB prediction
image

My graph scale has gone a bit crazy on Loop Follow as it now has prediction line for the carbs on board and therefore it’s really hard to see any detail of the BGL and dosing treatments. Good to be able set the graph Y axis scale. Thanks

bjorkert:
## Summary

- Clamp prediction graph values to a range of 39–400 mg/dL to prevent extreme predictions (e.g. high COB) from blowing up the graph scale
- Applies to both Loop and OpenAPS (ZT, IOB, COB, UAM) predictions
- The min/max info display still shows unclamped values so the actual predicted range remains visible

Fixes #495

### DIFF
--- a/LoopFollow/Controllers/Nightscout/DeviceStatusLoop.swift
+++ b/LoopFollow/Controllers/Nightscout/DeviceStatusLoop.swift
@@ -72,11 +72,9 @@ extension MainViewController {
                     while i <= toLoad {
                         if i < prediction.count {
                             let sgvValue = Int(round(prediction[i]))
-                            // Skip values higher than 600
-                            if sgvValue <= 600 {
-                                let prediction = ShareGlucoseData(sgv: sgvValue, date: predictionTime, direction: "flat")
-                                predictionData.append(prediction)
-                            }
+                            let clampedValue = min(max(sgvValue, 39), 400)
+                            let prediction = ShareGlucoseData(sgv: clampedValue, date: predictionTime, direction: "flat")
+                            predictionData.append(prediction)
                             predictionTime += 300
                         }
                         i += 1
--- a/LoopFollow/Controllers/Nightscout/DeviceStatusLoop.swift
+++ b/LoopFollow/Controllers/Nightscout/DeviceStatusLoop.swift
@@ -72,7 +72,7 @@ extension MainViewController {
                     while i <= toLoad {
                         if i < prediction.count {
                             let sgvValue = Int(round(prediction[i]))
-                            let clampedValue = min(max(sgvValue, 39), 400)
+                            let clampedValue = min(max(sgvValue, globalVariables.minDisplayGlucose), globalVariables.maxDisplayGlucose)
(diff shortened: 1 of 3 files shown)


marionbarker (commented):
I would really like the clamped values to be configured as named values. For example.

CLAMP_MAX_PREDICTION = 400
CLAMP_MIN_PREDICTION = 39

Configure those in a constants file and then use them. Try to get away from having hard-coded values that need to be modified, sometimes in multiple places,  if we ever decide to make a change.

This is not a requirement for this PR, but a suggestion moving forward.

marionbarker:
## Test

✅ Mostly does as expected.
Trio: I set up an outlier with a glucose value of 500 mg/dL so having the prediction flat-top at 400 doesn't look quite right.
Loop: there was not glucose outlier so prediction max and glucose tick-mark max were both visible at 400.

Do we want similar limits for outlier values for glucose.  For example if someone fat fingers 1000 into Apple Health instead of 100?

### Test Narrative

Uploaded a nasty pattern to Nightscout URL that includes 25 and 500.
Built this version and observe the scaling.

Started with glucose axis showing tic marks at 0, 50, 100, 150, 200, 250

The Glucose report is the actual value of 25 or 500, whereas Nightscout shows HIGH.

So this only clamps prediction values. The glucose value of 500 is displayed and glucose axis is now 0, 100, 200, 300, 400 and 500.

#### Trio Test

Go back to sine wave, change Trio CGM pattern to 160 +/- 100 with period of 1 hour, see if I can get some outlier predictions. Increase hours of prediction. Was 199/309 when 1 hour of prediction. Increase to 3 hours.

✅ For Trio, Min/Max shows: 166/431 for 3 hours, but glucose axis scales do not exceed the max glucose of 500.
❓ But also the plot flat tops at 400 when the glucose max range is 500 because that's the value

#### Loop Test

For Loop, don't need to play games with CGM. Can just add lots of carbs to get a high prediction.
Enter 100 g carbs, set a Manual Temp Basal to 0 U/hr for 2 hours.
Predictions is > 600 mg/dL
Increased hours of prediction to display to be 4 hours.

Initially, the glucose tic marks are 0, 50, 100, 150, 200, 250

dev version (without this PR): After loop where prediction shows min/max 117/681, the tick marks show 0 to 600 in steps of 100.

✅ build this version on the same phone: tick marks show 0 to 400 in steps of 100 and the prediction curve flat-tops at 400.

marionbarker (approved):
I approve but posted a few questions to consider. Let me know if you want to make modifications or go ahead with this as is.

bjorkert:
Good call — the constants now live in `globalVariables.minDisplayGlucose` / `globalVariables.maxDisplayGlucose` in `Globals.swift`, coordinated with #565 so both the main-screen HIGH/LOW labels and this prediction clamp share a single source of truth. If we ever want to adjust the range (there's a related discussion on #565 about Libre), it's now one place to change.

marionbarker (approved):

PR #565, opened 2026-03-21
Display HIGH/LOW for out-of-range BG on main screen

### LINKED ISSUE #503: Feature Request: Display "Low" and "High" for out-of-range glucose values
Feature Request: Display "Low" and "High" for out-of-range glucose values

Context : 

My 6-year-old son’s school caregiver uses LoopFollow. Recently, my son’s pod was ripped off, and his blood glucose (BG) rose to 22.3 mmol/L. Because the app continued to display "22.3" every five minutes, the caregiver assumed the system was working normally and did not realize the value was "out of range" or that the sensor had capped out.

The Dexcom Follow app uses "Low" and "High" labels to signal that a value is dangerously out of range and that the sensor can no longer provide an exact numerical measurement. Implementing this in LoopFollow would help non-technical caregivers immediately recognize when action is needed.

It became evident to me while I was at the school replacing the POD and trying to explain how I determined remotely that something wrong was happen.

Thanks.

bjorkert:
## Summary
- Show "HIGH" when BG >= 400 and "LOW" when BG <= 39 on the main screen, matching Dexcom sensor range behavior
- Reduce font size (85pt → 65pt) when displaying HIGH/LOW so the text fits
- Rename `setBGTextColor` → `updateBGTextAppearance` to reflect its broader role

Closes #503

### DIFF
--- a/LoopFollow/Controllers/Nightscout/BGData.swift
+++ b/LoopFollow/Controllers/Nightscout/BGData.swift
@@ -242,9 +242,9 @@ extension MainViewController {
             // Set BGText with the latest BG value
             self.updateBGTextAppearance()
 
-            if latestBG <= 39 {
+            if latestBG <= globalVariables.minDisplayGlucose {
                 Observable.shared.bgText.value = "LOW"
-            } else if latestBG >= 400 {
+            } else if latestBG >= globalVariables.maxDisplayGlucose {
                 Observable.shared.bgText.value = "HIGH"
             } else {
                 Observable.shared.bgText.value = Localizer.toDisplayUnits(String(latestBG))
--- a/LoopFollow/Controllers/Nightscout/BGData.swift
+++ b/LoopFollow/Controllers/Nightscout/BGData.swift
@@ -236,14 +236,19 @@ extension MainViewController {
             let latestBG = entries[latestEntryIndex].sgv
             let priorBG = entries[latestEntryIndex - 1].sgv
             let deltaBG = latestBG - priorBG
-            let lastBGTime = entries[latestEntryIndex].date
 
             self.updateServerText(with: sourceName)
 
             // Set BGText with the latest BG value
-            self.setBGTextColor()
+            self.updateBGTextAppearance()
 
-            Observable.shared.bgText.value = Localizer.toDisplayUnits(String(latestBG))
+            if latestBG <= 39 {
+                Observable.shared.bgText.value = "LOW"
+            } else if latestBG >= 400 {
+                Observable.shared.bgText.value = "HIGH"
+            } else {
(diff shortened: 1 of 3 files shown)


codebymini (approved):
👍

LoopFollow/Controllers/Nightscout/BGData.swift
```diff
@@ -236,14 +236,19 @@ extension MainViewController {
...
-            let lastBGTime = entries[latestEntryIndex].date
 
             self.updateServerText(with: sourceName)
 
             // Set BGText with the latest BG value
-            self.setBGTextColor()
+            self.updateBGTextAppearance()
 
-            Observable.shared.bgText.value = Localizer.toDisplayUnits(String(latestBG))
+            if latestBG <= 39 {
```
codebymini: Should these limits (39, 400) be defined in a global file to be reused?
(with https://github.com/loopandlearn/LoopFollow/pull/562)

LoopFollow/ViewControllers/MainViewController.swift
```diff
@@ -1105,6 +1105,12 @@ class MainViewController: UIViewController, UITableViewDataSource, ChartViewDele
             }
 
             BGText.textColor = color
+
+            if latestBG <= 39 || latestBG >= 400 {
```
codebymini: This could also be global values

LoopFollow/Controllers/Nightscout/BGData.swift
marionbarker: Please use named constants - see comments in PR #562

Also - do we really want to limit Libre output - which I think can go to 500. Also - some finger stick meters can go higher.

bjorkert: Thanks both. The thresholds are now in `globalVariables.minDisplayGlucose` / `globalVariables.maxDisplayGlucose` (0e86e9c) so future changes live in one place.

On the upper bound — fair point about Libre and meters. I've kept it at 400 for this PR so it stays consistent with the prediction clamp in #562, but I agree it's worth revisiting. Two ways we could handle it cleanly later:

1. Make the threshold user-configurable (e.g. in Graph Settings), so Libre users can bump it.
2. Only substitute HIGH/LOW for CGM readings and let manual BG entries through untouched.

The HIGH/LOW labels still reflect reality — they're just dropping the exact digits — but either of the options above would give users who want the precise number a path to it.

LoopFollow/ViewControllers/MainViewController.swift
bjorkert: Done — using `globalVariables.minDisplayGlucose` / `globalVariables.maxDisplayGlucose` from `Globals.swift` (0e86e9c).

marionbarker (approved):

PR #558, opened 2026-03-20
Units selection

codebymini:
## Summary
This PR centralizes glucose unit and metric preferences, adds a reusable units onboarding flow, and updates stats and setup screens to use the new configuration model. It also documents commit message conventions in the README.

## What changed
- Added a centralized UnitSettingsStore for glucose units and metric display preferences
- Introduced reusable units configuration UI and a dedicated onboarding screen
- Updated Nightscout and Dexcom setup flows to support units onboarding
- Updated stats views and formatters to use the new unit and metric settings
- Added support for exporting the new unit and metric preferences
- Documented commit message guidelines and best practices in the README
- Refactor Carbs screen to align with design from Loop App
- Poll NIghtscout data when remote commands been sent to populate the main graph as soon as possible

Screenshot 2026-03-20 at 14 32 35
Screenshot 2026-03-20 at 14 32 56
Screenshot 2026-03-20 at 14 33 03

PR #566, opened 2026-03-21
Add aggressive Nightscout polling after remote commands

bjorkert:
## Summary
- After a remote command (bolus, carbs, override, temp target) succeeds, poll Nightscout every 3s for up to 30s so the result appears on the main screen quickly
- Polling stops early once fresh data matching the command type is detected
- All remote command paths (Loop APNS, TRC, Trio Nightscout) post a `remoteCommandSucceeded` notification on success

Extracted from #558 to keep that PR focused on units/metrics.


bjorkert:
We might want to put this on hold, I am looking into socket communication with Nightscout that will give LF almost real time updates instead of polling, with polling as fallback.

bjorkert:
This PR is not needed since LF now has real time data using sockets.

PR #569, opened 2026-03-21
Deduplicate Nightscout treatment entries

bjorkert:
## Summary
- Nightscout can return duplicate treatment entries with the same `id` (Trio/Loop UUID) but different `_id` (MongoDB ObjectId), causing duplicate boluses, carbs, etc. to appear on the graph.
- Deduplicates entries by `id` at the top of `updateTreatments(entries:)` before classification, so all treatment types benefit. Entries without an `id` field are kept as-is.


marionbarker:
## Test

Split decision

* ✅ build this onto a SE 3rd gen phone following Trio site - no issues
* ❌ build this onto a SE 2nd gen phone following Loop and it never opens the display
   * switch that phone build to dev branch (6.0.10) - same issue
   * this is the phone on which I was running integration/dev-plus-prs
   * build that version and the display is fine.
   * Conclusion - there is something in the integration/dev-plus-prs branch that is not backward compatible
 
Delete LoopFollow from phone
✅ Build fix/deduplicate-treatments fresh on SE 2nd gen phone, configure to view Loop site and all is fine

marionbarker (approved):
approve by code review and test build

PR #571, opened 2026-03-22
Add Future Carbs Alert

bjorkert:
## Summary

- Add a new alarm type that fires once when a future-dated carb entry's scheduled time arrives, reminding the user to start eating
- Fix Nightscout treatment download query excluding future-dated entries
- Default the alarm to acknowledge mode since it self-silences when the carb is no longer in the future
- Add 10 unit tests for `FutureCarbsCondition` covering tracking, firing, deletion, lookahead bounds, min grams, stale cleanup, multi-carb behavior, and duplicate prevention

### DIFF
--- a/LoopFollow/Alarm/AlarmCondition/FutureCarbsCondition.swift
+++ b/LoopFollow/Alarm/AlarmCondition/FutureCarbsCondition.swift
@@ -6,13 +6,15 @@ import Foundation
 /// Fires once when a future-dated carb entry's scheduled time arrives.
 ///
 /// **How it works:**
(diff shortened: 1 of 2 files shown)


marionbarker:
## Test

Updated to commit 21e65f8.

Repeat test of future carbs of 10 g in 15 min.
The alarm is set to ignore future carbs > 10 minutes in the future.

Wait 15 minutes plus a few.

✅ I can see the future carbs in the LoopFollow plot
✅ No alert was given (as desired).

More testing soon.

PR #579, opened 2026-03-28
Add configurable CGM sensor lifetime for sensor change alarm

### LINKED ISSUE #583: Update for G7 15 day
Currently, it's hard-coded for Dexcom G7 to be 10 days and does not adapt if the person is wearing a new Dexcom 15-day sensor.

bjorkert:
## Summary

- Add a configurable sensor lifetime setting (7–15 days, default 10) to the Sensor Change Alert, so users with 15-day G7 sensors no longer get false early alerts
- Add unit tests for SensorAgeCondition covering both 10-day and 15-day lifetimes plus edge cases
Simulator Screenshot - iPhone 17 Pro Max - 2026-03-28 at 20 34 26


codebymini:
LGTM (Except the conflicts)

codebymini (approved):
👍

bjorkert:
Merge conflicts with `dev` resolved.

LoopFollow/Alarm/Alarm.swift
```diff
@@ -318,6 +327,7 @@ struct Alarm: Identifiable, Codable, Equatable {
         case .sensorChange:
             soundFile = .wakeUpWillYou
             threshold = 12
+            sensorLifetimeDays = 10
```
marionbarker, line 330: Should this remember what it was set last time the alarm was instantiated?

I set up a 15-day alarm.
Then I deleted the alarm.
I started a new alarm and it defaulted to 10-day.
Perhaps that's just the way it is.

LoopFollow/Alarm/AlarmCondition/SensorAgeCondition.swift
```diff
@@ -4,23 +4,22 @@
 import Foundation
 
 /// Fires once when we are **≤ threshold hours** away from the
-/// Dexcom 10-day hard-stop.  No repeats once triggered.
+/// sensor's configured lifetime.  No repeats once triggered.
```
marionbarker: This comment does not appear to be true. I got a 5-hour default snooze.

marionbarker (commented):
I had a few questions.

LoopFollow/Alarm/Alarm.swift
bjorkert: The alarm is currently self-contained and does not persist any state outside of the instance itself. That means creating a new alarm will always use the default value, even if a previous alarm was configured differently. This also allows multiple alarms to have different values, which is desirable for most settings, though arguably less useful in this specific case.

If we want this to “remember” the last used value, sensorLifetimeDays would need to live in shared settings rather than on the individual alarm.

The current approach was chosen since it keeps the setting close to where it is used.

LoopFollow/Alarm/AlarmCondition/SensorAgeCondition.swift
bjorkert: Good catch, I updated the comment.

marionbarker (approved):
Approved with updated comments and repeated test.
LGTM

PR #587, opened 2026-04-01
Fix carb data display in chart by updating marker line formatting

codebymini:
Fix label when touching carbs from Loop data

Before fix:
Screenshot 2026-04-01 at 14 09 16

After fix:
Screenshot 2026-04-01 at 14 09 01


bjorkert (approved):

marionbarker (approved):
Thanks for the graphics in the PR.

PR #589, opened 2026-04-05
Fix TRC bolus command not received by Trio

rafaeldelrey:
## Summary

- Switch TRC push notifications from `apns-push-type: alert` (priority 10) to `apns-push-type: background` (priority 5)
- Remove `alert` and `interruption-level` fields from the APNS payload

## Problem

Remote bolus commands sent via Trio Remote Control appeared to succeed (APNS returned 200) but Trio never received or acted on them. Trio's logs showed no trace of an incoming push notification.

The root cause: when `apns-push-type: alert` is used, iOS silently drops the notification if the receiving app (Trio) has not been granted notification permissions by the user. The command never reached `AppDelegate.didReceiveRemoteNotification`.

## Fix

Background push notifications (`content-available: 1`, `apns-push-type: background`, priority 5) are delivered by iOS regardless of notification permission status, which is the correct mechanism for remote command delivery. This matches Apple's documentation requirements — priority 5 is mandatory when using `apns-push-type: background`.

## Test plan

- [x] Send a remote bolus from LoopFollow (TRC) to a Trio device that has notification permissions denied
- [x] Confirm Trio receives and processes the command
- [x] Confirm APNS still returns 200
- [x] Verify other TRC commands (temp target, meal, override) also work

### DIFF
--- a/LoopFollow/Remote/TRC/PushMessage.swift
+++ b/LoopFollow/Remote/TRC/PushMessage.swift
@@ -7,20 +7,16 @@ struct EncryptedPushMessage: Encodable {
     let aps: APSPayload
     let encryptedData: String
 
-    init(encryptedData: String, commandType: TRCCommandType) {
+    init(encryptedData: String, commandType _: TRCCommandType) {
         self.encryptedData = encryptedData
-        aps = APSPayload(alert: "Remote Command: \(commandType.displayName)")
+        aps = APSPayload()
     }
 
     struct APSPayload: Encodable {
         let contentAvailable: Int = 1
-        let interruptionLevel: String = "time-sensitive"
-        let alert: String
 
         enum CodingKeys: String, CodingKey {
             case contentAvailable = "content-available"
-            case interruptionLevel = "interruption-level"
-            case alert
         }
     }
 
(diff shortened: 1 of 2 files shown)


LoopFollow/Remote/TRC/PushMessage.swift
```diff
@@ -7,20 +7,16 @@ struct EncryptedPushMessage: Encodable {
     let aps: APSPayload
     let encryptedData: String
 
-    init(encryptedData: String, commandType: TRCCommandType) {
+    init(encryptedData: String, commandType _: TRCCommandType) {
         self.encryptedData = encryptedData
-        aps = APSPayload(alert: "Remote Command: \(commandType.displayName)")
+        aps = APSPayload()
```
rafaeldelrey, line 12: This article says background notifications should not contain alert:  [developer.apple.com/documentation/usernotifications/pushing-background-updates-to-your-app](https://developer.apple.com/documentation/usernotifications/pushing-background-updates-to-your-app#Create-a-background-notification)

bjorkert:
Thanks for your interest in this.

Trio Remote Commands was originally implemented as a plain background push message, but due to the nature of how those work it was far from reliable — it works in good conditions, but otherwise the message might be delayed or not delivered at all. User reports were that too many messages didn't get through in time. In a recent change we moved to the current implementation (which also aligns with Loop/Loop Caregiver), and the success rate is much higher — user feedback supports this.

The case I think you're really getting at — messages not arriving when notifications are off for Trio — is better handled with a PR towards Trio itself: when remote commands are enabled, detect if notifications are disabled and require them to be turned on. If the notifications are disturbing, the sound option can be turned off for Trio notifications in iPhone Settings.

I'm closing this one, but feel free to comment anyway.

PR #537, opened 2026-03-12
Live activity

bjorkert:
## Summary

Adds Live Activity support to LoopFollow — a persistent lock screen and Dynamic Island widget that displays real-time glucose data, trend arrows, and key diabetes metrics without opening the app.

## New Features

- **Lock screen Live Activity** showing current glucose, trend arrow, delta, and a configurable grid of additional data (IOB, COB, loop status, pump/sensor age, etc.)
- **Dynamic Island** with redesigned compact view (BG + delta) and expanded view (large BG, trend, delta, projected glucose)
- **Configurable grid slots** — choose which InfoType values appear in each grid position via Live Activity settings
- **CarPlay Dashboard & Apple Watch Smart Stack** support via `.supplementalActivityFamilies` (iOS 18+)
- **SmallFamilyView** with a configurable right slot for the small activity family
- **APNs self-push updates** — the app sends push notifications to itself to keep the Live Activity updated in the background
- **Auto-renewal** — works around the iOS 8-hour Live Activity limit by automatically restarting the activity, with a user-facing renewal overlay when the app needs to come to the foreground
- **Audio session recovery** — detects and recovers from `AVAudioSession` failures, alerting the user via the Live Activity overlay
- **Deep-link tap navigation** — tapping the Live Activity (`loopfollow://la-tap`) navigates to the relevant tab in the app
- **Siri Intent** (`RestartLiveActivityIntent`) for restarting the Live Activity from Shortcuts or the system
- **Separate settings screens** — new Live Activity settings screen for layout/slot configuration, APN credentials remain in their own screen
- **Background refresh** via `BGAppRefreshTask` for improved background update reliability

## Infrastructure Changes

- New widget extension target (`LoopFollowLAExtension`) with its own entitlements and Info.plist
- App Group shared storage (`GlucoseSnapshotStore`) for passing glucose snapshots to the extension
- `SceneDelegate` added for scene-based URL handling (`scene(_:openURLContexts:)`)
- Mac Catalyst build guards (`#if !targetEnvironment(macCatalyst)`) to prevent compilation errors on macOS
- **CryptoKit migration** — replaced SwiftJWT/CryptoSwift with native CryptoKit for JWT signing in both `JWTManager` and `SecureMessenger`
- **JWTManager thread-safety** — added `NSLock` to prevent concurrent cache corruption when Live Activity pushes and remote commands race on different threads
- Live Activity termination handling (`endOnTerminate()`) that blocks up to 3s on app exit
- Fastlane updated for the new extension target

[...]

PR #595, opened 2026-04-09
Fix dev build: re-declare lastBGTime local in viewUpdateNSBG

bjorkert:
## Summary

- Re-adds `let lastBGTime = entries[latestEntryIndex].date` in `BGData.swift` `viewUpdateNSBG`. Dev currently fails to compile with `Cannot find 'lastBGTime' in scope`.

## Why this happened

Two PRs merged into `dev` on the same day (2026-04-09) within ~28 minutes, textually non-conflicting but semantically interfering:

- **#537 Live activity** (06:58 PDT) added `Storage.shared.lastBgReadingTimeSeconds.value = lastBGTime` later in the function, relying on the existing local `let lastBGTime`.
- **#565 Display HIGH/LOW for out-of-range BG on main screen** (07:26 PDT) removed the `let lastBGTime = entries[latestEntryIndex].date` declaration as part of an unrelated refactor — the #565 branch was cut before #537 landed, so from its perspective the local was unused.

Both PRs modified different regions of `viewUpdateNSBG`, so git's 3-way merge produced a syntactically clean file — but the surviving reference now dangles. Neither PR's CI caught it because each was built against an older base.

## Impact

`lastBgReadingTimeSeconds` has exactly one writer (this broken line) and one reader (`StorageCurrentGlucoseStateProvider.updatedAt`, which drives the Live Activity / Watch complication "updated at" timestamp). Restoring the write is the minimal fix.

### DIFF
--- a/LoopFollow/Controllers/Nightscout/BGData.swift
+++ b/LoopFollow/Controllers/Nightscout/BGData.swift
@@ -236,6 +236,7 @@ extension MainViewController {
             let latestBG = entries[latestEntryIndex].sgv
             let priorBG = entries[latestEntryIndex - 1].sgv
             let deltaBG = latestBG - priorBG
+            let lastBGTime = entries[latestEntryIndex].date
 
             self.updateServerText(with: sourceName)
 


marionbarker (approved):
Thanks for the fix.

PR #600, opened 2026-04-10
Clamp BG graph to HIGH/LOW display range

bjorkert:
## Summary

When a Nightscout reading exceeds `maxDisplayGlucose` (400 mg/dL), the main header correctly shows **HIGH** but the chart plots the raw value (e.g. 550) and autoscales the y-axis up to 600. This makes the chart inconsistent with the header text.

This PR clamps the plotted y-value in `updateBGGraph` to `[minDisplayGlucose, maxDisplayGlucose]` (39–400 mg/dL) so the chart line stays within the range the header represents. The pill tooltip still shows the raw reading so the exact value is accessible on tap.

- Chart-only change — `bgData`, stats, alarms, and color coding are unaffected.
- Predictions were already clamped to these bounds upstream (`DeviceStatusLoop`, `DeviceStatusOpenAPS`), so the graph is now consistent across both data types.
- Updated the `Globals.swift` comment to reflect the broader usage of the constants.

### DIFF
--- a/LoopFollow/Controllers/Graphs.swift
+++ b/LoopFollow/Controllers/Graphs.swift
@@ -801,10 +801,14 @@ extension MainViewController {
 
         topBG = Storage.shared.minBGScale.value
         for i in 0 ..< entries.count {
-            if Double(entries[i].sgv) > topBG - maxBGOffset {
-                topBG = Double(entries[i].sgv) + maxBGOffset
+            // Clamp the plotted y-value to the same bounds the header text uses
+            // (HIGH/LOW), so the graph stays consistent with the main display.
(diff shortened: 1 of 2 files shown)


bjorkert:
image
image
image
image

marionbarker (approved):

test: upload 550 and then 25 to Nightscout and use that site as CGM for a Loop test phone that is monitored by LoopFollow.

Expect the values 500 and 25 will be reported as High and Low. The y axis for the plot will be limited to 0 to 400 because of this PR. When you touch the dots, the information shows as 500 or 25 but the dots are placed at the 39 and 400 y-axis positions.

PR #601, opened 2026-04-11
Exclude Live Activity code from Mac Catalyst builds

bjorkert:
## Summary
- Wrap all Live Activity references behind `#if !targetEnvironment(macCatalyst)` compiler directives to fix build errors when targeting Mac Catalyst, since ActivityKit is not available on that platform.
- Affected files: LiveActivityManager, StorageCurrentGlucoseStateProvider, LiveActivitySettingsView, SettingsMenuView, MainViewController, SceneDelegate, RestartLiveActivityIntent.


marionbarker (approved):

Test with dev branch - could not build to catalyst
Test with fix/mac-catalyst-live-activity - no problem with build.

PR #604, opened 2026-04-12
Pin Send button to prevent keyboard toolbar overlap

### LINKED ISSUE #602: Trio Remote Control bolus UI - “Send bolus” button blocked when recommendation is present and value has been edited.
When 1) recommended bolus amount is present and 2) the bolus amount has been manually edited, the bin/hide keyboard buttons block the “Send bolus” button requiring you to either hide the keyboard or scroll the screen before you can click the send Bolus button. 
There appears to be padding at both top and bottom of screen that if reduced would allow enough space for the “Send bolus” button not to be blocked. 

Tested on iPhone 15 Pro Max, iOS 26.4. 

Image

bjorkert:
## Summary

- Moves the Send button out of the scrollable Form and pins it at the bottom of the screen using `.safeAreaInset(edge: .bottom)` in all four remote command views

The issue reports that the Send button gets blocked by the keyboard toolbar (trash/dismiss buttons) when a recommended bolus is present and the value has been edited. Rather than reducing padding as suggested in the issue, the Send button is now pinned outside the Form entirely — it stays visible above the keyboard at all times, regardless of form content height or screen size.

**Affected views:** TRC Bolus, TRC Meal, Loop APNS Bolus, Loop APNS Carbs

Fixes #602


bjorkert:
image
image

marionbarker (approved):

Tested TRC meal and bolus, Loop carbs and bolus using an SE phone.

Button was nicer and more consistent, no scrolling required with this version compared to version before this PR.

While testing the older version for comparison, the PR branch was updated by merging dev into the branch. No change to PR diffs, but test again anyway. Very nice.

LGTM

PR #605, opened 2026-04-12
Add cone of uncertainty for Trio predictions

bjorkert:
## Summary

- Adds a "cone of uncertainty" visualization for Trio predictions, matching Trio's native forecast display. Instead of four individual prediction lines (ZT, IOB, COB, UAM), a filled band shows the min/max envelope across all prediction arrays at each 5-minute time step.
- Adds a Prediction Style setting (Cone/Lines) for Trio users. Loop users are locked to its single prediction line with the picker hidden.
- Reactively monitors device type changes via Combine to auto-set the appropriate default (cone for Trio, lines for Loop) while preserving manual user preference across Nightscout refreshes.

image


marionbarker (approved):
Approve by test.

**Loop** phone - no difference in LoopFollow, Settings, Graph in the PREDICTION section.

**Trio** phone - new row, `Prediction Style` in LoopFollow, Settings, Graph in the PREDICTION section.
* The default setting is Cone - I guess that will get people's attention when they build this plus it matches the Trio default
* Modify setting to Lines for both LoopFollow and the Trio phone it is following. The lines visually look the same
* Restore to cone and displays are consistent in the two apps.

PR #606, opened 2026-04-13
Add Nightscout WebSocket support for real-time data updates

bjorkert:
## Summary

- Adds optional Socket.IO WebSocket connection to Nightscout for real-time push notifications of data changes
- WebSocket delta events selectively trigger HTTP fetches only for changed data types (BG, treatments, device status, profile), reducing unnecessary network requests
- Polling intervals are extended (3x for BG, 5x for treatments/device status, 3x for profile) while WebSocket is authenticated, with immediate fallback to normal polling if the connection drops
- WebSocket is foreground-only: it disconnects when the app moves to the background and reconnects on return, so the persistent connection only runs while you're using the app
- Toggle in Nightscout settings defaults to on; info sheet describes the foreground/background behavior

## Companion PR

Documentation: [loopandlearn/loopfollowdocs#32](https://github.com/loopandlearn/loopfollowdocs/pull/32)


bjorkert:
image
image

bjorkert:
Addressed both review findings:

**Stale WebSocket session on config change (high):** `updateStatusLabel()` now calls `disconnect()` when URL/token validation fails, so the old socket can't keep streaming from the previous server. Additionally, `removeAllHandlers()` is now called before `disconnect()` in the teardown path — this prevents intentional disconnects from firing the `.disconnect` event handler, which would post a refresh and potentially reconnect to the invalid URL.

**Staleness timer (medium):** Removed entirely. The extended polling intervals (3x for BG/device status, 5x for treatments, 3x for profile) already continue running as a safety net. If the socket authenticates but no data flows, polling still fetches via HTTP at the extended cadence. The staleness timer was redundant and had a bug where it never armed if no `dataUpdate` was received.

codebymini (approved):
👍  LGTM

marionbarker (approved):

Issue #593, opened 2026-04-09
feature bug: Live Activity display has truncation problems


marionbarker:
## Problem

The new Live Activity feature, currently found in dev, has some truncation issues.

This issue includes a screenshot of one such issue.

![Image](https://github.com/user-attachments/assets/fe46c0cd-d026-408e-9544-d396fc73c9c3)

For this case the scheduled basal of 0.65 U/hr is visible but the full string: "0.65 -> 0" is truncated so the currently applied Temporary Basal Rate is not visible.

marionbarker:
We had similar issues with Loop.

This PR was added to assist in truncation for larger fonts.
* Loop PR 2397: [Modify live activity](https://github.com/LoopKit/Loop/pull/2397#top)

bjorkert:
Would it make sense to only show the current/effective basal in this view? Given the space constraints, that might be clearer than trying to display both values and risking truncation.

MtlPhil:
Given the space constraint, the idea that we only show effective basal on the Live Activity makes sense.

I'll first play around with the size. I think we can increase each cell’s width (e.g., 60 → 72) and allow the value text to shrink more aggressively (minimumScaleFactor ~0.65 + allowsTightening(true)), while slightly reducing the left column width to free space. 

If this preserves the aligned 2×2 matrix but gives each slot enough flexibility to avoid truncation, I'll keep it. Otherwise, I'll fall back on only showing the effective basal.

I don't have an iPhone SE I can test the visual on, though. I can test on an iPhone 17, which is 1.6 inches wider.

MtlPhil:
![image](https://github.com/user-attachments/assets/fe6db966-4a32-47ac-aab5-512ab08106a7)

Will test with longer basal string. If it doesn't work on a smaller screen, let's do only effective basal.

marionbarker:
closed with merge of #610

PR #610, opened 2026-04-19
Adjust frame sizes and text scaling in LiveActivity

MtlPhil:
Issue #593


marionbarker (approved):

PR #613, opened 2026-04-25
Add App Group capability to Fastfile

marionbarker:
With the Live Activity feature, an App Group is required for LoopFollow.

Add this capability to Fastfile so that users just need to configure the App Group when preparing Identifiers.

While updating the Fastfile, remove some unused lines as well.


dnzxy (approved):

bjorkert (approved):
LGTM

PR #615, opened 2026-04-26
Fix Live Activity restart classification, foreground race, and add troubleshooting logs

bjorkert:
## Summary

Two fixes plus diagnostics for Live Activity restart paths, split out from the broader background-renewal work so they can land independently.

### Restart classification + foreground race
- `handleExpiredToken`, `endOnTerminate`, `forceRestart`: mark `endingForRestart` before ending so the state observer doesn't misclassify the resulting `.dismissed` as a user swipe (which would set `dismissedByUser=true` and block auto-restart on the next background refresh).
- Defer foreground restart from `willEnterForeground` to `didBecomeActive` so `Activity.request()` isn't called before the scene is active (avoids the "visibility" failure).
- Remove duplicate orphan `LiveActivitySettingsView.swift` under `Settings/` (not referenced by the Xcode project).

### Troubleshooting logs
- `startIfNeeded`: log entry state (authorized, activities, current, flags) and enrich `Activity.request` failure with `NSError` domain/code + scene state.
- `renewIfNeeded`: enrich catch with `NSError` domain/code + authorization state.
- `handleForeground` / `handleDidBecomeActive`: include `applicationState` and the existing activities count at entry.
- `observePushToken`: log token fingerprint (last 8 chars) and prior value so token rotations are visible.
- `update`: log when the direct ActivityKit update is skipped (app backgrounded) and when APNs is skipped because no push token has been received yet.
- `performRefresh`: log the gate that blocks LA updates — especially `dismissedByUser=true`, which previously caused silent extended outages.
- `handleExpiredToken`: log current id, activities count, and flags before ending so APNs 410/404 events are correlatable to the restart path.
- `bind`: include `activityState` and the previous `endingForRestart` value so the dismissal-classification path is traceable.


marionbarker (approved):
Approve by test that the app builds and LiveActivity shows on the watch and phone.

This has been used successfully by the developer on his personal phone.

PR #618, opened 2026-04-27
Remove the Nightscout-based remote command path

bjorkert:
## Summary
- Trio dev's APNS-based remote command support has merged to main, so the Nightscout treatment-posting variant of remote commands is no longer needed.
- Drops the `.nightscout` case from `RemoteType`, deletes `TrioNightscoutRemoteView.swift`, `TrioNightscoutRemoteController.swift`, and the now-orphaned `NoRemoteView.swift` fallback, and removes every `.nightscout` switch arm in remote settings, import/export, and device validation paths.
- Storage keys are reused by the TRC and Loop APNS variants and remain untouched — no schema changes.

## Migration
Existing users with `remoteType` previously set to "Nightscout" automatically fall back to `.none` on next launch: `StorageValue`'s `JSONDecoder` fails on the removed enum case and returns the default value. No explicit migration step is needed.


marionbarker (approved):

PR #622, opened 2026-04-27
Add iOS 17.2+ push-to-start for Live Activity renewal

bjorkert:
## Summary

On iOS 17.2 and later, the Live Activity now renews itself silently in the background instead of asking you to open the app every 7–8 hours. iOS 16 keeps the existing behavior unchanged.

## How it works

Apple ends every Live Activity after about 8 hours. Until now, when that limit got close, LoopFollow had to wait for you to bring the app to the foreground so it could request a fresh one — and if you didn't, you'd get a notification asking you to.

iOS 17.2 added a way to start a Live Activity from a server-side push instead of from the foreground app. LoopFollow now uses that:

- When it's time to renew, the app sends an APNs push to itself. iOS creates a new Live Activity from the push, and the old one is replaced silently.
- The same push-to-start path is used when you toggle the Live Activity on or hit Restart, so there's only one creation flow on iOS 17.2+.
- Updates still ride APNs the same way they did before.

Because everything goes through APNs, valid APNs credentials become a hard requirement for Live Activities, to make that obvious before something silently breaks, the Settings UI now validates them.

## Settings UX

- **APN page**: shows a green check or orange warning next to the Key ID and APNs key as you type, with a one-line hint if either looks wrong.
- **Live Activity page**: the section footer points to the APN settings, and a warning row appears above the Restart button if the credentials are missing or invalid.

## iOS 16 behavior

Unchanged. Same `Activity.request()` flow, same renewal-failed notification when the app is backgrounded at the deadline. The iOS 17 paths are gated by `#available` so the legacy code can be removed in one go when the deployment target eventually moves to 17.2.


marionbarker (commented):
After merging PR 615 to dev - there are conflicts with this PR.

When I test this branch stand-alone, it builds fine and I see LA on phone and watch.

Once conflicts with dev, commit 06b314f, are resolved, I will test again.

marionbarker (approved):
Approve by test build and because developer has been using this code personally.

* Test on real hardware: SE 3rd gen running iOS 26.4.2
* Test on real hardware: SE 2nd gen running iOS 18.7.8
* Test on real hardware: iPhone 15 running iOS 26.4.1
* Mac Catalyst

As a proof of the Issue, test dev (without this PR) on a phone using Silent Tunes as background refresh and enable Live Activity.

* Observe that while the phone was locked, the Live Activity values did not update.
* During that same time, observe a phone running this PR with APN entered kept the Live Activity values up to date with the Nightscout site.

PR #623, opened 2026-04-27
Redact secrets from log output

bjorkert:
## Summary

Logs can contain bits of information that some users would rather not share. To make people more comfortable sharing logs when asking for help, this change masks sensitive data before they're written to the file, while keeping the lines readable enough to actually debug from.

## How it works

A new `LogRedactor` helper provides small functions for the common shapes:

- `tail` / `head` — keep last/first 8 chars of an opaque token
- `url` — keep scheme + a redacted host hint, drop path and query. Known managed Nightscout hosts (`nightscoutpro.com`, `10be.de`, `herokuapp.com`) are kept as the suffix; anything else is reduced to the TLD
- `keyId` / `teamId` — reveal last 2 chars of a 10-char Apple ID
- `bundleId` — keep TLD + app name, mask middle (so suffixes like `.watchkitapp` stay visible)
- `username` — preserve email-vs-not shape, drop the value
- `fingerprint` — short sha256 prefix for opaque blobs (QR contents, settings JSON)
- `sweep` — runs on every log line as a safety net, catching PEM PRIVATE KEY blocks, `?token=…` query values, and JWTs no matter which call site emits them

The per-call-site helpers are precise; the central `sweep` in `LogManager.log` is a net so future log lines also benefit without anyone having to remember.

## What changed

Per-call-site redaction at the known sites that previously echoed values verbatim:

- `AppDelegate` — APNs device token at registration; `userInfo` dumps reduced to key lists
- `NightScoutViewController` — `NSLog` of the full URL routed through `LogManager` and masked
- `MainViewController` — Nightscout URL in the BFU foreground log
- `ImportExportSettingsViewModel`, `RemoteSettingsViewModel` — QR / settings JSON replaced with a fingerprint; `nightscoutURL` and `dexcomUsername` masked
- `LoopAPNSService` — `bundleId` masked next to the (already prefix-masked) device token
- `JWTManager` — Apple Key ID in the JWT generation log
- `PushNotificationManager`, `NightscoutUtils`, `SettingsMigrationManager` — `print`s converted to debug `LogManager` calls (no raw JSON, no full `debugDescription`)
- `ObservableUserDefaultsValue`, `ObservableValue` — removed `print`s that fired on every value change

### DIFF
--- a/LoopFollow.xcodeproj/project.pbxproj
+++ b/LoopFollow.xcodeproj/project.pbxproj
@@ -248,6 +248,7 @@
 		DDD10F0B2C54192A00D76A8E /* TemporaryTarget.swift in Sources */ = {isa = PBXBuildFile; fileRef = DDD10F0A2C54192A00D76A8E /* TemporaryTarget.swift */; };
 		DDDB86F12DF7223C00AADDAC /* DeleteAlarmSection.swift in Sources */ = {isa = PBXBuildFile; fileRef = DDDB86F02DF7223C00AADDAC /* DeleteAlarmSection.swift */; };
 		DDDC01DD2E244B3100D9975C /* JWTManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = DDDC01DC2E244B3100D9975C /* JWTManager.swift */; };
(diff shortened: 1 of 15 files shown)


marionbarker (approved):
I think this is pretty important.
I tested it and all new log entries were redacted but older ones were not.
When the log was shared - the older entries were in the clear.

I'm going to merge this and we can decide if filtering needs to be added when sharing the logs. I'm thinking not. The logs have a finite life so unmasked entries will be short-lived.

PR #627, opened 2026-04-29
Default debug logging on and prompt for description when sharing logs

bjorkert:
## Summary

- Flip the `debugLogLevel` default from `false` to `true` so fresh installs get useful logs out of the box. Add `migrateStep8` that force-enables it for existing users whose stored value is still `false`.
- Replace the silent Share Logs flow with a small prompt: a SwiftUI sheet asks the user to describe the problem (when it happened, what they did, what they expected). The text is written to `ShareNotice_<timestamp>.txt` (with date, app version, and branch+sha header) and bundled into the iOS share sheet alongside today's and yesterday's log files.
- An empty description is allowed; the notice file then reads `(no description provided)`.

## Companion PR

Documentation: [loopandlearn/loopfollowdocs#31](https://github.com/loopandlearn/loopfollowdocs/pull/31)


codebymini (approved):
LGTM

marionbarker (approved):

Sharing state (one enabled and one disabled was not modified by building).
Build on a simulator to demonstrate default button is to accept sharing.

marionbarker:
My previous comment did not make sense.
This is enabling log default on with share (different kind of share.)
When I checked Advanced, the debug log was enabled for both test phones.
At any rate - going with the developers approval for this and merging.

PR #628, opened 2026-04-29
Fix duplicate chevron on Import/Export Settings row

bjorkert:
## Summary
The Import/Export Settings row in Remote Settings used a `NavigationLink` (which renders its own chevron) and also added a manual `chevron.right` image, producing a doubled `>>`. Removed the manual chevron so the row matches the rest of the app.


bjorkert:
Before:
image
After:
image

marionbarker (approved):

- See the double chevron with dev
- See single chevron with this PR

PR #631, opened 2026-05-01
Default migrationStep to latest so fresh installs skip migrations

bjorkert:
## Summary

Previously `migrationStep` defaulted to `0`, forcing every fresh install to iterate through every migration step. The migrations are designed for upgraders carrying legacy state — for a brand-new install they're all no-ops or set values that the static defaults could just hold directly.

This change sets the default `migrationStep` to the current latest step (7) and updates the other `StorageValue` defaults so a fresh install ends up in the same state as a user who ran every migration:

- `snoozerPosition`: `.menu` → `.position3`
- `nightscoutPosition`: `.position3` → `.position4`
- `remotePosition`: `.position4` → `.menu`
- `hasSeenFatProteinOrderChange`: `false` → `true` (matches what step 4 sets for fresh installs via `!isExistingUser`)

`alarmsPosition` (`.position2`) and `homePosition` (`.position1`) already matched the post-migration state.

Reminder comments are added at the `migrationStep` declaration and at the `runMigrationsIfNeeded()` block so future authors remember to bump the default and update affected defaults whenever a new migration step is added.

Existing users are unaffected: they have a stored `migrationStep` value and continue to run only the steps they haven't yet executed.

### DIFF
--- a/LoopFollow/Storage/Storage.swift
+++ b/LoopFollow/Storage/Storage.swift
@@ -33,7 +33,7 @@ class Storage {
     var mealWithFatProtein = StorageValue<Bool>(key: "mealWithFatProtein", defaultValue: false)
 
     // TODO: This flag can be deleted in March 2027. Check the commit for other places to cleanup.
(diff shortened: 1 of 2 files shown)


marionbarker (approved):
Approve from code review and test build.

PR #633, opened 2026-05-03
Recognize Atlas DASH pod in Omnipod heartbeat scan

### LINKED ISSUE #630: Background Refresh: scan does not find Atlas pods (Omnipod Dash)
When a user selects *Background Refresh → Omnipod Dash*, scanning never finds the pod if it is an **Atlas** pod (newer Insulet hardware revision). Reported by a user who is looping with Dash.

The matcher in `BackgroundRefreshType.matches(_:)` for `.omnipodDash` only accepts peripherals whose advertised name is exactly `"TWI BOARD"` (a real Dash pod) or `" :: Fake POD ::"` (the rPi DASH simulator):

```swift
case .omnipodDash:
    if let name = device.name {
        // actual DASH or rPi DASH simulator
        return name == "TWI BOARD" || name == " :: Fake POD ::"
    }
    return false
```
*LoopFollow/BackgroundRefresh/BackgroundRefreshType.swift:62*

`OmnipodDashHeartbeatBluetoothTransmitter` connects on service UUID `1A7E4024-E3ED-4464-8B7E-751E03D0DC5F` and reads characteristic `1A7E2442-...`, which is the standard (non-Atlas) Dash pod service.
*LoopFollow/BackgroundRefresh/BT/Devices/OmnipodDashHeartbeatBluetoothTransmitter.swift:8*

Because Atlas pods use a different advertised name (and likely a different service UUID), the BLE scan list never offers them and the *Omnipod Dash* heartbeat option is effectively unavailable to Atlas users.

## What to do

- Identify the advertised name and BLE service UUID of an Atlas pod (a packet capture from a phone connected to an Atlas pod is enough).
- Either:
  - extend `BackgroundRefreshType.matches(_:)` and `OmnipodDashHeartbeatBluetoothTransmitter` to also accept the Atlas advertisement and service, or
  - add a separate `.omnipodAtlas` case if the timing/characteristics differ enough to need their own transmitter class.

Until then, Atlas users have to fall back on Silent Tune or a Dexcom heartbeat for background refresh.

bjorkert:
## Summary
- Fixes #630: Atlas (newer Insulet) DASH pods were never offered by the Bluetooth scan when *Background Refresh → Omnipod Dash* was selected.
- Atlas pods advertise the BLE peripheral name `"InPlay BLE"` instead of `"TWI BOARD"`, but expose the same OmniBLE service (`1A7E4024-…`) and data characteristic (`1A7E2442-…`), so the existing `OmnipodDashHeartbeatBluetoothTransmitter` works as-is — only the scanner's name matcher needed to accept the new name.

### DIFF
--- a/LoopFollow/BackgroundRefresh/BackgroundRefreshType.swift
+++ b/LoopFollow/BackgroundRefresh/BackgroundRefreshType.swift
@@ -61,8 +61,10 @@ enum BackgroundRefreshType: String, Codable, CaseIterable {
 
         case .omnipodDash:
             if let name = device.name {
-                // actual DASH or rPi DASH simulator
-                return name == "TWI BOARD" || name == " :: Fake POD ::"
+                // "TWI BOARD": original DASH pod
+                // "InPlay BLE": newer Atlas DASH pod
+                // " :: Fake POD ::": rPi DASH simulator
+                return name == "TWI BOARD" || name == "InPlay BLE" || name == " :: Fake POD ::"
             }
             return false
 


bjorkert:
Verified with real hardware.
BBDA8B7F-1B60-43CE-A5D7-88D6043FF3ED_1_102_o

marionbarker (approved):
Approve from Code Review.

This was tested on real hardware.
Confirmed this branch builds as expected.

PR #635, opened 2026-05-04
Diagnose and skip rogue Nightscout profile records

bjorkert:
## Summary

- Switches the Nightscout profile fetch from `/profile/current` to `/api/v1/profiles?count=1&find[startDate][$lte]=now`, so future-dated records can no longer hide the active profile. `/profile/current` only sorts by `startDate` desc and ignores whether that date is in the past or future, which makes a single rogue upload (e.g. a phone with the wrong system clock) keep winning indefinitely. The `find[]` filter is the same query syntax already used for treatments, sage/iage/cage, BG data, and stats.
- Adds a "Run diagnostics" button in the Remote Settings → Debug section. It fetches up to 1000 recent profile records and surfaces three known failure modes for remote commands:
  - **Profile uploaded by a different app** — when Loop and Trio share a Nightscout, profile uploads overwrite each other. Detected by checking which key path carries `bundleIdentifier` (Loop populates `loopSettings.bundleIdentifier`, Trio populates the top-level field) against `Storage.shared.device.value`.
  - **Multiple devices uploading profiles** — multiple installs (spare/old phones) each push their own deviceToken, causing APNS commands to fail intermittently. Detected by compressing consecutive same-token runs and warning when a token reappears (true alternation), not when tokens simply rotate (`a→b→c` stays silent). The warning lists each shift with timestamp and abbreviated from→to tokens, so the user can see which devices are competing.
  - **Future-dated profile record found** — informational only, since the new fetch already skips them; flagged because the rogue record still corrupts what Nightscout's own dashboard considers current and signals a broken phone clock somewhere.
- Profile records are sorted by `startDate` with a fallback to `created_at`, so uploaders that omit `startDate` don't cluster at the start and corrupt the bouncing-tokens compression.
- Results auto-scroll into view when the run completes (`ScrollViewReader` anchored to the diagnostics block).
- Reuses existing helpers — `NightscoutUtils.executeRequest`/`parseDate`, the `find[]` parameter convention, `dateTimeUtils.formattedDate(from:)`, the orange `exclamationmark.triangle` warning idiom from `ImportExportSettingsView`. No new abstractions.

### DIFF
--- a/LoopFollow.xcodeproj/project.pbxproj
+++ b/LoopFollow.xcodeproj/project.pbxproj
@@ -109,6 +109,7 @@
 		DD4878052C7B2C970048F05C /* Storage.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD4878042C7B2C970048F05C /* Storage.swift */; };
 		DD4878082C7B30BF0048F05C /* RemoteSettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD4878072C7B30BF0048F05C /* RemoteSettingsView.swift */; };
 		DD48780A2C7B30D40048F05C /* RemoteSettingsViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD4878092C7B30D40048F05C /* RemoteSettingsViewModel.swift */; };
+		AB1CD0012C7B30D40048F05C /* RemoteDiagnostics.swift in Sources */ = {isa = PBXBuildFile; fileRef = AB1CD0022C7B30D40048F05C /* RemoteDiagnostics.swift */; };
 		DD48780E2C7B74A40048F05C /* TrioRemoteControlViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD48780D2C7B74A40048F05C /* TrioRemoteControlViewModel.swift */; };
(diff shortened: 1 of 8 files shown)


bjorkert:
Image

bjorkert:
6DA3B5E9-4D3A-46F2-80F3-A05700E672CA

marionbarker:
## Test

Build this version on my test phone

### Test Trio and Loop both uploading

* Configure LoopFollow to look at my Trio test Nightscout site
* Make sure a Trio app is actively uploading to the site
* Configure a Loop app to upload to the same site
   * Modify therapy settings on the Loop app to force a profile update
* on LoopFollow phone, tap on Remote Settings, 
   * even though I previously selected Trio Remote Control, I needed to select it again
   * scroll to the bottom and run diagnotics
   * ✅ get Profile uploaded by a different app message 

### Test two versions of Loop uploading

* Configure LoopFollow to look at my Loop test Nightscout site, change Remote to be Loop Remote Control
* One Loop test phone was already uploading to that site
* Configure a second Loop test phone to also upload and then modify a therapy setting
* on LoopFollow phone, tap on Remote Settings, 
   * scroll to the bottom and run diagnotics
   * ✅ get multiple devices uploading profiles warning

### Test future time

I don't want to do this test because it screws up the test phone.  I'll examine the code instead.

marionbarker (approved):

PR #638, opened 2026-05-06
Suppress silent LA notifications and prevent stale foreground-restart

bjorkert:
## Summary

Three independent fixes for Live Activity behaviour that surfaced together
in a 6.1.0 report (audible notification on LA renewal + a gratuitous LA
restart 37 minutes after a successful background renewal).

- **`willPresent` no longer forces banner+sound for every notification.**
  The handler in `AppDelegate` returned `[.banner, .sound, .badge]`
  unconditionally, which meant any notification iOS routed through
  `UNUserNotificationCenter` while LF was foregrounded would produce
  sound — including the intentionally-silent push-to-start payload
  (`interruption-level: passive`, empty title/body). The handler now
  returns `[]` for passive notifications and for ones with empty
  title/body. The four intentional alerts (`scheduleRenewalFailed…`,
  `scheduleApnsCredentialsMissing…`, `schedulePushToStartTokenMissing…`,
  alarms) all use non-empty title/body and the default `.active`
  interruption level, so they continue to surface normally.

- **`pendingForegroundRestart` is cleared on every push-to-start
  adoption** (`adoptPushToStartActivity`). Without this, a brief
  foreground entry while the renewal overlay was up could latch the
  intent, the user could background again before `didBecomeActive` ran,
  the background renewal could replace the LA, and the next foreground
  entry — minutes later — would tear down the freshly-renewed LA and
  start another.

- **`performForegroundRestart` re-checks the latch conditions.** Belt
  and braces with the previous fix: even if a future code path latches
  the intent without going through adoption, the deferred restart will
  bail out when `renewalFailed`/`overlayShowing`/`pushToStartLooksStuck`
  no longer hold.

- **Deferred-foreground-restart push-to-start is tagged
  `reason="deferred-foreground-restart"`** instead of inheriting the
  `"user-start"` label from `startIfNeeded`. The previous tag made the
  stale-latch event indistinguishable from a real user action in logs.
  Threaded through a single-shot `nextStartReasonOverride` so the
  override is consumed exactly once and `startIfNeeded`'s public surface
  is unchanged.

- **`willPresent` log line expanded** with `interruptionLevel` and
  whether title/body are empty, so future reports can confirm whether
  iOS is routing the LA push-to-start payload through `willPresent`.

### DIFF
--- a/LoopFollow/Application/AppDelegate.swift
+++ b/LoopFollow/Application/AppDelegate.swift
@@ -231,12 +231,21 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
                                 willPresent notification: UNNotification,
                                 withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void)
     {
-        // Log the notification
-        let userInfo = notification.request.content.userInfo
-        let userInfoKeys = userInfo.keys.compactMap { $0 as? String }.sorted()
-        LogManager.shared.log(category: .general, message: "Will present notification: keys=\(userInfoKeys)")
+        let content = notification.request.content
+        let userInfoKeys = content.userInfo.keys.compactMap { $0 as? String }.sorted()
+        LogManager.shared.log(
+            category: .general,
+            message: "Will present notification: keys=\(userInfoKeys), interruption=\(content.interruptionLevel.rawValue), title=\(content.title.isEmpty ? "empty" : "set"), body=\(content.body.isEmpty ? "empty" : "set")"
+        )
 
-        // Show the notification even when app is in foreground
+        // Suppress notifications iOS routes here that we never intended to surface:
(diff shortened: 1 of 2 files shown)


MtlPhil:
If you're making fixes to the LA, you may want to take a look at these two quick hits as well, in PR 639.

[https://github.com/loopandlearn/LoopFollow/pull/639](url)

bjorkert:
Splitting this PR — opening three logically scoped follow-ups so each problem can be reviewed in isolation:

- willPresent-only fix: #NEW1
- Spurious foreground-restart latch fix: #NEW2
- 410 recovery (new, addresses a separately reported issue): #NEW3

Will edit this comment with the actual numbers once they open.

bjorkert:
Replaced by:
- willPresent only: #655
- Foreground-restart latch: #656
- 410 recovery (separate issue from a follow-up log report): #657

PR #647, opened 2026-05-08
Larger text in Snoozer display (more accessible)

aug0211:
- Increase sizing of text on snoozer display (delta, time since reading, etc)
- Respect safety limits and "snoozer" popup bar
- Friendly down to iPhone 13 Mini size display
- Consistent spacing (no text jitter) on snoozer display thanks to consistent formatting and .monospacedDigit() use


bjorkert:
Thanks for this, the bigger text is a win for the night stand use case and I have taken it forward in #749, which keeps the font sizes, the full white delta and reading age, and the monospaced digits from here.

I left out the min ago and scheduler changes. The min ago text is deliberately calm: it updates once a minute, and only switches to minutes and seconds between 4.5 and 12 minutes after the last reading, when a new reading is due and it helps to see how long is left. That text is also shared with the home screen.

Two things in the layout part turned out to be regressions. The early return in scheduleAutoHide reads the alarm before the published change lands, so the snoozer bar never hides after snoozing an alarm. And moving the snooze card into the scaled column made it shrink and float off the bottom in landscape. In #749 the card stays in the overlay and the auto hide logic is untouched.

I also found on a 13 mini that the bigger rows squeezed the glucose value smaller than before when the emoji was on, so #749 gives the reading layout priority and the emoji and clock a minimum height. Closing this one in favour of #749.

PR #648, opened 2026-05-09
Add Trio smoothed BG display option

aug0211:
- Adds Display Smoothed BG toggle in Settings → Advanced (off by default).
- Pulls smoothed values from Nightscout devicestatus openaps.enacted.bg / openaps.suggested.bg and stores an in-memory history bounded by the graph's day range.
- Renders smoothed values as a light-grey line on the main BG chart; CGM dots stay colorful, but the dot-connecting line is hidden while smoothing is on. Display Dots / Display Lines toggles are disabled in Graph Settings during this mode.
- Chart-tap popup shows the smoothed value (✨ … ✨) above the raw CGM value when a match is available within tolerance of the dot's timestamp.
- All work (fetches, parsing, polling) is gated on the toggle so users who don't enable it incur no extra overhead.
- Add the option for users to display the smoothed value in the information display table as well.


dnzxy:
Good idea to bring this in for LF x Trio.

bjorkert:
Merged `dev` in to resolve conflicts, plus two small fixes:

- Restored the `LF_DEVELOPMENT_TEAM` build setting in the project file (it had been blanked out).
- Resolved a dataset-index collision with dev's new "yesterday" line: `smoothedBg` is now index 20 (yesterday keeps 19), with a matching placeholder added to the small chart so both charts stay aligned.
- Changed the smoothed BG line from grey to cyan so it doesn't clash with the grey yesterday overlay (footer text updated to match).

PR #650, opened 2026-05-09
Bug fix: carb dots overlapping SMB dots on the chart

aug0211:
- Restores pre-PR #305 behavior; no other dot positioning changes
- Carb dots entered alongside an SMB were rendering at the same Y position as the SMB triangle, visually overlapping
- The carb-vs-bolus collision logic was supposed to lift the carb dot above the bolus to BG+70, but only checked for Boluses (not SMB)
- Regression comes from PR #305 (SMB display differentiated from boluses), which split SMBs into their own smbData array. - After the split, findNearestBolusbyTime in processNSCarbs no longer saw SMBs as "nearby treatments," so carb-near-SMB fell through to the default BG+20 offset: the same Y as the SMB itself.
- processNSCarbs now also looks up smbData.
- A carb within 300s of either a bolus OR SMB lifts to BG+70; otherwise BG+20.

### DIFF
--- a/LoopFollow/Controllers/Nightscout/Treatments/Carbs.swift
+++ b/LoopFollow/Controllers/Nightscout/Treatments/Carbs.swift
@@ -10,6 +10,7 @@ extension MainViewController {
         carbData.removeAll()
         var lastFoundIndex = 0
         var lastFoundBolus = 0
+        var lastFoundSmb = 0
 
         for currentEntry in entries.reversed() {
             var carbDate: String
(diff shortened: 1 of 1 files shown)


bjorkert (approved):
Verified on top of the Swift Charts migration: with a carb next to an SMB the carb now lifts to BG+70 as intended, carbs next to manual boluses still lift, and lone carbs stay at BG+20. Merges cleanly into current dev. Thanks for the fix and the clear regression analysis!

marionbarker (approved):
Approved by test using Trio URL.

Applied patch to dev v6.2.25.  added carb, observed smb. the two labels are offset and readable.

PR #651, opened 2026-05-09
Add tappable zoom buttons to main graph view

aug0211:
- Add tappable buttons to main view for zooming in/out/selecting pre-defined zoom level
- Very nice to have on iOS, but also fantastic for users deploying on Mac, where pinch to zoom is not supported
- Disable pinch to zoom on Mac to avoid force closes


bjorkert:
I can see the value of this, possibly the buttons should only be shown on Mac? The branch has drifted a lot though, and includes unrelated changes. Could you open a fresh one from current dev with just the zoom buttons?

aug0211:
Odd yeah, somewhere along the way a lot of other changes got pushed here.

PR #652, opened 2026-05-10
Bugfix: phantom ZT prediction dot intercepting latest BG dot on graph…

aug0211:
… tap

- Bugfix: tapping latest BG dot sometimes showed a stale ZT prediction popup rather than the actual CGM's BG reading
- Fix: refresh predictions on BG update; drop entries within ~1 min of latest BG


bjorkert:
Thanks @aug0211, and sorry this sat so long. Closing it now that the Swift Charts migration (#704) has removed Controllers/Graphs.swift entirely.. in the new chart, prediction marks aren't tappable at all, so the phantom ZT popup can no longer occur.

PR #608, opened 2026-04-18
Remove storyboard, migrate to SwiftUI layout

bjorkert:
## Summary

Removes Main.storyboard and SceneDelegate, replacing the entire UI lifecycle with SwiftUI. The home screen layout (BG display, info table, charts, stats) is now composed in a single SwiftUI view (`MainHomeView`) with the DGCharts `LineChartView` instances embedded via `UIViewRepresentable`.

**What changed:**

- Storyboard and SceneDelegate deleted, replaced by `LoopFollowApp` (SwiftUI App) and `MainTabView`
- All five tab view controllers (Alarms, Settings, Snoozer, Remote, More Menu) replaced with SwiftUI views
- Info table migrated from `UITableView` to SwiftUI `List` backed by `InfoManager` as an `ObservableObject`
- Statistics area and pie chart migrated to SwiftUI (pie chart still uses DGCharts `PieChartView` via `UIViewRepresentable` since the Charts pod shadows Swift Charts)
- BG display area (BG value, direction, delta, min ago, loop status, prediction) migrated from UILabels to SwiftUI with pull-to-refresh via `.refreshable`
- Main layout rewritten from `UIStackView` hierarchy to `MainHomeView` SwiftUI composition — visibility of info table, small graph, and stats is now reactive via Storage observables
- `getMainViewController()` hierarchy-walking methods replaced with `MainViewController.shared` weak reference
- Cleaned up dead code: unused imports, stale `tabBarController` references, redundant appearance notification relay, unused `UIViewExtension`

**Net result:** −1430 lines, all UIKit view controllers except `MainViewController` (charts) and `NightscoutViewController` (web view) are gone.

PR #655, opened 2026-05-12
Suppress silent passive notifications in willPresent handler

bjorkert:
## Summary

`UNUserNotificationCenterDelegate.willPresent` in `AppDelegate` returned
`[.banner, .sound, .badge]` unconditionally, so any notification iOS
routed through the handler while the app was foregrounded produced
sound — including the Live Activity push-to-start payload, which is
intentionally silent (`interruption-level: passive`, empty title/body).
6.1.0 users reported the silent renewal triggering an audible banner.

## Change

- Return `[]` for passive notifications and for notifications with
  empty title/body.
- The four intentional alerts (`scheduleRenewalFailed…`,
  `scheduleApnsCredentialsMissing…`, `schedulePushToStartTokenMissing…`,
  alarms) all use non-empty title/body and the default `.active`
  interruption level, so they continue to surface as banners.
- Expanded the `willPresent` log line with `interruptionLevel` and
  whether title/body are empty so future reports can confirm whether iOS
  is routing a given payload through this handler.

Replaces #638 (split into three independently reviewable PRs).

### DIFF
--- a/LoopFollow/Application/AppDelegate.swift
+++ b/LoopFollow/Application/AppDelegate.swift
@@ -231,12 +231,21 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
                                 willPresent notification: UNNotification,
                                 withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void)
     {
-        // Log the notification
-        let userInfo = notification.request.content.userInfo
-        let userInfoKeys = userInfo.keys.compactMap { $0 as? String }.sorted()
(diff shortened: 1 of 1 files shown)


codebymini (approved):
👍

marionbarker (approved):
I had some trouble building this because it came after we merged #608 into dev, but does not include 608 in this branch.  Did the trick figured out during 608 testing to build successfully. (get blank screen, quit and restart app).

* Test for Loop URL
* Test for Trio URL
* Test for iOS 18.7.8 and iOS 26.5

PR #656, opened 2026-05-12
Prevent stale foreground-restart latch from tearing down freshly renewed LA

bjorkert:
## Summary

A 6.1.0 report showed a gratuitous Live Activity restart ~37 minutes
after a successful background renewal. `pendingForegroundRestart` could
outlive the condition that triggered it: a brief foreground entry while
the renewal overlay was up latched the intent, the user backgrounded
before `didBecomeActive` ran, the background renewal then replaced the
LA, and the next foreground entry minutes later fired the deferred
restart against an already-fresh LA.

## Changes

- **`adoptPushToStartActivity` clears `pendingForegroundRestart` on
  every adoption.** A freshly-adopted LA resolves the renewal-window
  condition that latched the intent, so the deferred restart would be
  needless work.
- **`performForegroundRestart` re-checks the latch conditions**
  (`renewalFailed`, `overlayShowing`, `pushToStartLooksStuck`) and bails
  if none still hold. Belt-and-braces in case a future path latches the
  intent without going through adoption.
- **Deferred-foreground-restart push-to-start is tagged
  `reason="deferred-foreground-restart"`** via a single-shot
  `nextStartReasonOverride` consumed in `startIfNeeded`. The previous
  `"user-start"` label made the stale-latch event indistinguishable from
  a real user action in logs.

Replaces #638 (split into three independently reviewable PRs).

### DIFF
--- a/LoopFollow/LiveActivity/LiveActivityManager.swift
+++ b/LoopFollow/LiveActivity/LiveActivityManager.swift
@@ -174,6 +174,17 @@ final class LiveActivityManager {
         Storage.shared.laRenewalFailed.value = false
         cancelRenewalFailedNotification()
         dismissedByUser = false
+        // A fresh LA invalidates any latched foreground-restart intent — the
+        // condition that prompted the latch (overlay showing / renewal failed)
+        // is resolved by adoption itself, so a deferred restart on the next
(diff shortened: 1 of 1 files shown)


marionbarker (approved):
Tested that it builds and Live Activity works.

PR #657, opened 2026-05-12
Recover Live Activity after APNs 410 token expiry

bjorkert:
## Summary

A 6.1.0 user reported the Live Activity vanishing overnight and
refusing to come back without a manual Restart. Log trace:

- 04:12 — scheduled renewal succeeds, new LA adopted.
- 04:42 — `APNs token expired (410) — restarting Live Activity`,
  `handleExpiredToken` ends the activity.
- 04:42 – 08:42 — `activities=1, current=nil, renewBy=0.0`; every
  foreground entry logs `no action needed (not in renewal window)`.
  No new LA is created.
- 08:42 — `.dismissed: endingForRestart=false, renewalFailed=false,
  pastDeadline=false, renewBy=0.0` → classified as **user swipe**,
  `dismissedByUser=true` latches. From here every refresh logs
  `LA update skipped — dismissedByUser=true` until the user opens
  the app and taps Restart.

## Root cause

Two cooperating bugs around an app-initiated `end()`:

1. `end()` nulls `current` and clears `laRenewBy`. `handleExpiredToken`'s
   comment said the LA would restart on the next BG refresh via
   `refreshFromCurrentState()`, but `renewIfNeeded` short-circuits when
   `current` is nil, and `performRefresh`'s `bind-existing` path then
   rebinds to the just-ended activity (still listed in
   `Activity<>.activities` until iOS purges it). `bind()` clears
   `endingForRestart`, so the late `.dismissed` reads as `renewBy=0` /
   `renewalFailed=false` / `endingForRestart=false` — branch (c) "USER"
   in the classifier.

2. The classifier had no way to recognize a stale observer firing for an
   activity the app no longer tracks.

## Fixes

- **`handleExpiredToken` drives the restart synchronously on iOS 17.2+**
  (`attemptPushToStartCreate(reason: "expired-token")`), so the orphaned
  post-410 state is short-lived and adoption of the fresh activity
  cancels the old observer via `attachStateObserver`.
- **`performRefresh` / `update` bind-existing only to `.active`
  activities.** Binding to an `.ended`/`.dismissed` corpse would clear
  `endingForRestart` and re-attach an observer that only ever delivers
  `.dismissed`.
- **`.dismissed` classifier gains branch (d): stale observer.** If the
  dismissed activity is not the one we currently track, log and take no
  action. Only the foreground LA can be user-swiped, so a stale-observer
  delivery for an already-replaced activity must not latch
  `dismissedByUser=true`. The `wasCurrentActivity` signal is captured
  before `current` is cleared so legitimate user swipes still classify
  correctly.

### DIFF
--- a/LoopFollow/LiveActivity/LiveActivityManager.swift
+++ b/LoopFollow/LiveActivity/LiveActivityManager.swift
@@ -1093,7 +1093,13 @@ final class LiveActivityManager {
             LogManager.shared.log(category: .general, message: "[LA] refresh: LA update skipped — areActivitiesEnabled=false reason=\(reason)")
             return
         }
-        if current == nil, let existing = Activity<GlucoseLiveActivityAttributes>.activities.first {
+        if current == nil,
+           let existing = Activity<GlucoseLiveActivityAttributes>.activities.first(where: { $0.activityState == .active })
+        {
+            // Skip activities already in .ended/.dismissed — those are corpses
+            // (typically post-410 ends pending iOS dismissal). Binding to them
+            // would clear endingForRestart and turn the eventual iOS dismissal
+            // into a misclassified user swipe.
             bind(to: existing, logReason: "bind-existing")
         }
         if let _ = current {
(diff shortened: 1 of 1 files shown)


marionbarker (approved):
Tested that it builds and Live Activity works.

PR #661, opened 2026-05-24
Deduplicate Dexcom Share readings to fix delta always zero

bjorkert:
## Summary

- Dexcom Share returns each reading twice when both the iPhone Dexcom app and the Apple Watch app upload to the same Share account. The two copies arrive ~9–10 seconds apart with identical SGV values.
- Without deduplication the two most recent entries in `bgData` were always identical, so `delta = latestBG - priorBG = 0` every time.
- The Nightscout fetch path already had inline deduplication (30-second / same-SGV filter). Extracted it into a shared helper `deduplicateBGReadings(_:)` and applied it to the Dexcom-only path as well.


marionbarker (approved):
Approved from code review. Test that build works.

PR #662, opened 2026-05-28
Fix BG history truncation caused by duplicate readings from multiple uploaders

bjorkert:
When multiple apps upload BG data to the same account — e.g. a closed-loop system plus the G7 iPhone app plus Apple Watch — duplicate readings are written to both Nightscout and Dexcom's cloud. This causes 2–3 entries per 5-minute slot instead of 1, which silently truncates history in two places:

**Nightscout fetch:** the entry count cap (`days × 2 × 288`) was too small. With 3 uploaders producing ~700 entries/day, the cap of 576 was hit after ~20h, dropping the oldest ~4h. Raised the multiplier from `2` to `4` in both `BGData.swift` and `StatsDataFetcher.swift`, giving headroom for up to 4 uploaders. The `date[$gte]` filter still bounds the window so no extra data is returned when fewer entries exist.

**Dexcom Share fetch:** the Share API has a hard cap of 288 readings regardless of `maxCount`. With 2 readings per slot, those 288 cover only ~15h instead of 24h. Two fixes:
- Deduplicate Dex Share data with the same 30-second window used for Nightscout, so the graph receives clean readings.
- When the deduplicated Dex data doesn't reach back to the start of the requested window, fall through to `webLoadNSBGData(dexData:)` so Nightscout fills in the older portion using the existing merge logic.


codebymini (approved):
👍 LGTM

marionbarker (approved):

Issue #15, opened 2020-06-14
Dex Share


jonfawcett:
For those times when NS is delayed with a reading, I'd like to have the option to automatically pull from Dex Share.

If BGtime > 5 min, automatically do a dex share refresh for however many BG readings we need.

jonfawcett:
Pushed to Dev.

- Dex is used by default if it is entered
- NS will be used if Dex fails
- NS will be checked if Dex is overdue
- All settings, views, and alerts automatically shift if using Dex only.

PR #663, opened 2026-05-28
Bump fastlane to 2.235.0 and jwt to 3.2.0 (CVE-2026-45363)

bjorkert:
Resolves Dependabot alert #15 (GHSA-c32j-vqhx-rx3x / CVE-2026-45363).

`jwt < 3.2.0` accepts attacker-forged tokens when an empty or nil key is supplied to HMAC algorithms (HS256/HS384/HS512). `fastlane 2.233.1` had a hard `jwt < 3` upper bound that blocked the fix. `fastlane 2.235.0` relaxes that to `jwt < 4`, allowing the upgrade to `jwt 3.2.0`.


marionbarker:
I think these 2 lines should be removed from Gemfile:
```
gem "json", ">=2.19.2"
gem "addressable", ">=2.9.0"
```
I added them when they were needed for security issues with fastlane 2.233.1.

When I bumped [LoopKit/LoopWorkspace up to fastlane 2.234.0](https://github.com/LoopKit/LoopWorkspace/pull/438/changes/0eac0f8f638), they were no longer needed so I removed them.

I tested `bundle install` for this branch after removing those 2 lines and the Gemfile.lock was modified only by removing those two lines in the DEPENDENCIES section.

bjorkert:
Good catch — removed both pins. `bundle install` confirms the only Gemfile.lock change is dropping those two DEPENDENCIES lines, resolved versions stay safe.

codebymini (approved):
👍 LGTM

marionbarker (approved):
Approve by code review and test Browser Build.
* https://github.com/docs-test/LoopFollow/actions/runs/27507761113

PR #665, opened 2026-05-30
Add optional yesterday BG comparison line to main graph

bjorkert:
Adds an optional comparison overlay that draws yesterday's BG curve on the main graph, time-shifted by 24h so it lines up with the same clock time today. It's a thin, dimmed gray "ghost" line (no dots, no tap popups) purely for visual comparison of how the day is tracking versus the day before.

## Details
- New **Show Yesterday's BG** toggle in Graph Settings → Graph Display (Nightscout only, default off). Dexcom Share only returns ~24h, so the 24–48h-ago window requires Nightscout.
- When enabled, the BG fetch window is extended by one extra day; the normal display window (`bgData`) is unchanged, while the full set — shifted +24h — feeds the overlay.
- The overlay is capped at *now + Hours of Prediction*, so it never extends further into the future than the prediction line.
- Does not affect today's BG line, axis scaling, scrolling, stats, alarms, or the small graph. The line pans and zooms together with the main graph and extends as the day progresses.

## Notes
- The shift is a fixed +24h, so on daylight-saving change days the overlay is off by one hour versus wall-clock time — acceptable for a comparison aid.


marionbarker (approved):
Test with iOS 18 and iOS 26 phones and using Loop and Trio URLs.
Show yesterday's BG works for both URLs.

PR #666, opened 2026-05-31
ITMS-91061 - privacy manifest for Charts framework

luborjurena:
App Store submission was rejected with ITMS-91061: Missing privacy manifest
for Frameworks/Charts.framework. The bundled Charts pod predates
Apple's privacy-manifest requirement and ships no PrivacyInfo.xcprivacy, so
the built framework lacks one.

This adds a privacy manifest to the Charts framework.

TF error:
ITMS-91061: Missing privacy manifest - Your app includes “Frameworks/Charts.framework/Charts”, which includes Charts, an SDK that was identified in the documentation as a commonly used third-party SDK. If a new app includes a commonly used third-party SDK, or an app update adds a new commonly used third-party SDK, the SDK must include a privacy manifest file. Please contact the provider of the SDK that includes this file to get an updated SDK version with a privacy manifest. For more details about this policy, including a list of SDKs that are required to include signatures and manifests, visit: https://developer.apple.com/support/third-party-SDK-requirements.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -33,4 +33,36 @@ post_install do |installer|
       File.write(transformer, code.sub(original, patched))
     end
   end
+
+  # Inject a privacy manifest into the Charts framework (ITMS-91061).
+  # Charts 4.1.0 ships no PrivacyInfo.xcprivacy; it collects no data, performs
+  # no tracking, and uses no required-reason APIs, so this is a negative
+  # declaration. Re-applied here because `pod install` regenerates the project.
+  charts_manifest = <<~XML
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+    <dict>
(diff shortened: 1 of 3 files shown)


bjorkert:
Thanks for digging into this, but I want to flag the underlying approach: LoopFollow isn't meant to be submitted to the App Store. The app does a number of things that wouldn't pass App Store review, so a public App Store submission isn't a supported distribution path and will keep running into rejections like this one.

The intended distribution is the **TestFlight track only** (via the GitHub browser build pipeline or a local archive uploaded to TestFlight), plus building directly to your device from Xcode. TestFlight builds don't go through App Store review, so the ITMS-91061 privacy-manifest rejection you hit shouldn't block a TestFlight distribution.

If you're seeing this error on a TestFlight upload specifically, let us know the details and we can look closer — but if this came from an actual App Store submission, that's the part to step away from.

luborjurena:
Yes, I know. I have external testers in a TestFlight group, and the build approval was rejected with the error mentioned above.

bjorkert:
Ok, thanks for the clarification.

bjorkert (approved):

marionbarker (approved):
Approved by @bjorkert 
Tested by building with Mac-Xcode and Browser Build.

PR #667, opened 2026-06-03
Scale info data table text with Dynamic Type

bjorkert:
## Summary

The top-right info data (IOB / COB / Basal / etc.) used a hardcoded font size and fixed row heights, so it ignored the iPhone's text-size / accessibility (Dynamic Type) setting.

This drives the info table's font size and row height from `@ScaledMetric`, so the text grows and shrinks with the user's text-size setting while keeping today's compact 17 pt / 21 pt look at the default size. Both the font and the row height scale together, so larger text never clips. The scaling is capped at `accessibility1` so the compact top strip (shared with the BG display, constrained to 250 pt wide) stays within its layout at extreme sizes.

Scope is intentionally limited to the info table.

Simulator Screenshot - iPhone 17 Pro Max - 2026-06-03 at 19 28 48

### Changes
- `LoopFollow/InfoTable/InfoTableView.swift` — `@ScaledMetric(relativeTo: .body)` for font size (17) and row height (21); applied to the font, the row frame height, and `defaultMinListRowHeight`.
- `LoopFollow/ViewControllers/MainHomeView.swift` — `.dynamicTypeSize(...DynamicTypeSize.accessibility1)` on the info table to cap the scaling.


marionbarker:
## Test

I set the font size to largest fonts and see this change between dev (left side) and info-table-dynamic-type (right side).
Even going half-way up the larger font size scale caused some truncation.

❓ Should the larger fonts be configured to use two rows each: left side first row is label (left justified), right side each row is result (right justified), both being auto-scaled to fit in available space?



lf-pr-673

bjorkert:
Good catch, give the latest commit a try: 
CleanShot 2026-06-14 at 22 22 57@2x

marionbarker:
Much better. For this example, however, the override name is pretty long. It was auto-named by Trio when adding an override and starting it without naming it. I think in that case, the font should shrink to fit the value in the second line if already using two lines. I enabled all possible values and all looked good except for the Override.

If you disagree, I'll merge as is.

IMG_7741

marionbarker:
## Test

✅ successful test

### Summary of modifications since last time
* If there is no value associated with a row, indicate by using the "-" character
* If the font is large enough to require two lines and the value overflows the second line, shrink to font to fit (as much as a factor of 2). 

### Screenshot

The graphic below shows default, mid-range and max-size text using accessibility sizes. The Information table scrolls up and down as needed.

lf-pr-667

marionbarker:
### Screenshot continued

I mean to include a screenshot with no override. This is with the largest possible font.

IMG_7232

marionbarker (approved):
Tested with iOS 18 and 26.
Tested with Loop and Trio URLs.

PR #668, opened 2026-06-03
Match Trio's override/temp-target colors for non-Loop backends

bjorkert:
Overrides and temp targets were drawn with the colors swapped relative to Trio. In Trio an override is purple and a temp target is green; LoopFollow had them reversed.

Now non-Loop backends (Trio, OpenAPS) follow Trio's scheme — override purple, temp target green — while Loop keeps its existing colors (override green, temp target purple). The choice keys off the existing `device == "Loop"` convention, so it reacts to a backend change without recreating the graph.

Also switches the override info-row population in `Overrides.swift` from `== "Trio"` to `!= "Loop"` so it applies to all non-Loop backends.

### DIFF
--- a/LoopFollow/Controllers/Graphs.swift
+++ b/LoopFollow/Controllers/Graphs.swift
@@ -7,6 +7,22 @@ import UIKit
 
 import Charts
 
+/// Fill colors for the override and temp-target bars on the BG graph.
+///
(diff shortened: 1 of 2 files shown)


codebymini (approved):
Approved by code review

marionbarker (approved):
Confirm the Trio lines on LoopFollow match Trio lines in Trio.

PR #675, opened 2026-06-17
Remove empty Nightscout group left over from #618

bjorkert:
## Summary
- #618 removed the Nightscout-based remote command path, deleting the only file under `LoopFollow/Remote/Nightscout/`.
- The now-empty `Nightscout` `PBXGroup` was left behind in `project.pbxproj`, still referencing a folder with no contents.
- This removes the orphaned group and its reference from the parent Remote group. No source or build changes.


marionbarker (approved):

PR #677, opened 2026-06-20
Prefer structured ISF/CR from device status over reason string

bjorkert:
When showing ISF and Carb Ratio in the info table, LoopFollow reads from the most recent `enacted`/`suggested` record in device status.

**ISF** was taken straight from the structured `ISF` field without validation. The value is now validated, and a `0` is treated as missing, so the info row falls back to the profile ISF instead of rendering `profileISF → 0`.

**Carb Ratio** was parsed out of the `reason` string by regex, even though device status already carries a structured `CR` field. CR now reads from the structured field when present (and non-zero), keeping the `reason`-string regex only as a fallback for uploaders that don't expose it. Parsing the structured number is more robust than matching free text across oref versions.


marionbarker (approved):
Approve from code review and minimal testing. It builds, it reports modified ISF for Trio URL.

PR #678, opened 2026-06-20
Fix stale snooze state in alarm list

bjorkert:
The alarm list sorted alarms into Snoozed vs. Active by comparing `snoozedUntil` against `Date()` at render time, but nothing re-rendered the view once a snooze expired. A lapsed snooze could sit in the Snoozed section for hours while the alarm editor correctly showed it as not snoozed.

Refresh a `now` snapshot on `onAppear` and when the app returns to the foreground, and route the snooze comparisons (both sections and the row label) through it.


marionbarker (approved):

PR #679, opened 2026-06-20
Fix tab bar labels wrapping at large Dynamic Type sizes

bjorkert:
At large accessibility text sizes the tab bar titles wrapped and truncated ("Re…", "Sno…") because both the title font and the SF Symbol icons scaled with Dynamic Type, leaving no room in the fixed-height tab bar.

This pins the title font (global `UITabBar` appearance) and the icons (fixed-size symbol images) to a constant size. Works regardless of which tabs the user has placed in the bar.


bjorkert:
When testing, the tab font did not scale.

PR #680, opened 2026-06-20
Extend predictive low alarm to Trio

bjorkert:
The low alarm's predictive look-ahead only handled Loop's single forecast. Trio publishes four separate forecasts (ZT, IOB, COB, UAM), so the predictive series was empty and the early low warning never fired.

This builds the alarm's forward series from the lowest of the four forecasts at each point in time, giving Trio an equivalent predictive-low signal: the alarm fires if any forecast dips to or below the threshold.

Note: the look-ahead was in practice broken for Loop as well. Since the 3.0 alarm rewrite, the BG-limit gate treated the low alarm's `belowBG` as an activation limit and blocked evaluation until the latest reading was already at or below it, so the early warning could never fire ahead of a low (this worked in 2.x). Conditions can now declare that they check their own threshold; `LowBGCondition` does, and evaluates the latest reading, persistence, and forecast in one place.

- `AlarmTask`: new `lowestForecast` combiner and `alarmPredictionData`, which feeds the low condition for both Loop and Trio.
- `AlarmCondition`: new `checksOwnBGLimits` so a condition can own its BG threshold instead of having it enforced as an activation limit.
- `LowBgAlarmEditor`: wording now refers to the forecast instead of Loop's prediction, and drops the "not available for Trio" note.
- Tests for the forecast combining and the low condition across Loop and Trio inputs, including `shouldFire`-level coverage of the gate.

Also adds a missing `import Foundation` to `SensorAgeConditionTests`, which otherwise failed to compile under the current Xcode.


marionbarker:
## Test

✅ successful test
❓ Do we want to indicate it is a predictive alert?

## Test with Loop URL

For LoopFollow tracking Loop URL:
* configure predictive low alert for LoopFollow Urgent Low alert
* configure Loop test phone to log an external dose of 10 U
* eventually updates to -443 mg/dL
* LoopFollow phone alarms for Urgent Low even though glucose is 114

There is no indication that this is a predictive alert.

## Test with Trio URL

For LoopFollow tracking Trio URL:
* configure predictive low alert for LoopFollow Low BG alert
* configure Trio test phone to log an external dose of 10 U
* eventually updates to -443 mg/dL
* LoopFollow phone alarms for Low BG even though glucose is 111

There is no indication that this is a predictive alert.

marionbarker (approved):
If we later decide to flag the reason an alert alarms (when predictive is enabled), that will be a different PR.

PR #681, opened 2026-06-20
Add search filter to the Add Alarm sheet

bjorkert:
Adds a search field to the Add Alarm sheet so the alarm list can be filtered while typing.

- Search matches against each alarm's name, blurb, and group.
- Group sections hide when they have no matching alarms.
- Shows a "No Results" state when nothing matches.


bjorkert:
Simulator 2026-06-20 18 10 39

marionbarker (approved):

PR #682, opened 2026-06-21
release.sh: verify latest script before running

bjorkert:
Guards the release flow against running an outdated copy of `release.sh`.

Before any git operations, the script now fetches `origin/dev` and compares the running script (`${BASH_SOURCE[0]}`) against `origin/dev:release.sh`. If they differ, it aborts with guidance to switch to the latest `dev` and re-run.

This catches the cases where a stale script would otherwise run:
- launched from an old branch
- on `dev` but behind `origin/dev`

The check is on script content, not branch — the release flow switches branches itself, so what matters is that the logic being executed is current.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -68,6 +68,18 @@ update_follower () {
 PRIMARY_ABS_PATH="$(pwd -P)"
 echo "🏁  Working in $PRIMARY_ABS_PATH …"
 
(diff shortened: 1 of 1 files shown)


marionbarker (approved):
Approve by code review.

PR #685, opened 2026-06-21
release.sh: set gh pr repo explicitly

bjorkert:
Makes the release flow's PR creation independent of `gh`'s inferred default repository.

When a clone has multiple remotes, `gh` cannot guess which GitHub repo to act on and `gh pr create` fails with "No default remote repository has been set" — after the pushes have already happened, leaving the release half-finished.

This derives `owner/repo` from the `origin` remote URL (handling both HTTPS and SSH forms) and passes `--repo` to both `gh pr create` calls.


marionbarker (approved):
Approve from code review.

PR #686, opened 2026-06-21
release.sh: drop [skip ci] from version bump; allow manual tag_on_main

bjorkert:
The `[skip ci]` on the release version-bump commit had two harmful side effects observed during the v6.2.0 release:

- It skipped the required **SwiftFormat** lint check, leaving the release PRs stuck at `BLOCKED` (requiring a manual override to merge).
- It skipped **tag_on_main.yml**, so the `v6.2.0` tag was never created automatically and had to be pushed by hand.

The `[skip ci]` was only there to stop `auto_version_dev` from re-bumping when the commit lands on `dev` — but that workflow already prevents this via its own "Skip if Config.xcconfig was changed in this push" guard. So the token is redundant for its intended purpose and is removed.

Also adds a `workflow_dispatch` trigger to `tag_on_main.yml` as a safety net: if a tagging run is ever skipped, the tag can be created by dispatching the workflow against `main` (idempotent — it skips if the tag already exists).

Note: once this is merged, the `SwiftFormat` requirement can safely be enabled for `main` (Phase 3) without blocking future release PRs.

### DIFF
--- a/.github/workflows/tag_on_main.yml
+++ b/.github/workflows/tag_on_main.yml
@@ -17,6 +17,10 @@ on:
   push:
     branches:
       - main
+  # Manual fallback: if the push-triggered run was ever skipped (e.g. a commit
+  # with [skip ci] landing on main), the tag can be created by dispatching this
+  # workflow against main. It is idempotent — it skips if the tag already exists.
+  workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


marionbarker (approved):
Approve from code review.

PR #688, opened 2026-06-22
Sanitize Nightscout token to prevent WebSocket crash on launch

bjorkert:
## Problem

LoopFollow crashes on startup (the reporting user has TestFlight 6.2.0, iPad on iOS 26.5) with an `EXC_BREAKPOINT`/SIGTRAP from `URLComponents.percentEncodedQuery`'s setter, reached through Socket.IO's `createURLs()` while opening the Nightscout WebSocket in `MainViewController.viewDidLoad`.

The Nightscout WebSocket (Socket.IO) is new in 6.2.0, so this is the first build that exercises this path. Socket.IO builds its connect URL query from the `token` we pass in `connectParams`, and its `urlEncode()` doesn't escape whitespace/control characters. On iOS 26 the `percentEncodedQuery` setter is strict and traps on a string that isn't already valid percent-encoding. So a token holding a stray character (e.g. a trailing newline from a paste) produces an invalid query and crashes the app at launch.

## Fix

- Add `NightscoutUtils.sanitizeToken(_:)`, which strips whitespace, newlines, and control characters.
- Sanitize the token when it's written from the Nightscout settings screen.
- Sanitize defensively at WebSocket connect time so already-saved bad tokens stop crashing without the user re-entering anything.

Verified by the user reporting the issue, it's no longer crashing on startup and works like it should

### DIFF
--- a/LoopFollow/Controllers/Nightscout/NightscoutSocketManager.swift
+++ b/LoopFollow/Controllers/Nightscout/NightscoutSocketManager.swift
@@ -40,10 +40,10 @@ class NightscoutSocketManager {
             return
         }
 
-        let url = Storage.shared.url.value
-        // Sanitize defensively: tokens saved before this fix may still hold a stray
+        // Sanitize defensively: values saved before this fix may still hold a stray
         // whitespace/control char that crashes Socket.IO's URL builder on iOS 26.
-        let token = NightscoutUtils.sanitizeToken(Storage.shared.token.value)
+        let url = NightscoutUtils.sanitizeConnectionInput(Storage.shared.url.value)
+        let token = NightscoutUtils.sanitizeConnectionInput(Storage.shared.token.value)
 
         guard !url.isEmpty else {
             disconnect()
--- a/LoopFollow/Controllers/Nightscout/NightscoutSocketManager.swift
+++ b/LoopFollow/Controllers/Nightscout/NightscoutSocketManager.swift
@@ -41,7 +41,9 @@ class NightscoutSocketManager {
         }
 
         let url = Storage.shared.url.value
-        let token = Storage.shared.token.value
+        // Sanitize defensively: tokens saved before this fix may still hold a stray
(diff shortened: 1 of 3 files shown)


marionbarker:
## Test

✅ This modification fixes the specific error of a token with white space
❌ This modification does not fix the entry of a URL with embedded token with white space

I think the URL entry and token entry both need to be sanitized.

I will approve this to avoid blocking the merge if this is urgent.
I would prefer this contain a more general fix.

### Configuration

iPhone SE 2nd gen running iOS 18.7.8
* build LoopFollow with fix/ws-token-sanitize-crash

### Test Details

#### Enter URL with Token

I entered a properly formatted URL with embedded token.
* Works fine.

I then copied that string and added both a space at the end and a return at the end for 2 different tests.
* In both instances, I got an Unknown error when pasted the modified string into the URL space for Nightscout.

#### Enter URL and Token Separately

If I separately input the URL and the token and enter a token with an extra space at the end, it correctly reads the token.

Repeat that test with dev and get invalid token.

marionbarker (dismissed):
Based on code review and test:
✅ This modification fixes the specific error of a token with white space
❌ This modification does not fix the entry of a URL with embedded token with white space

I think the URL entry and token entry both need to be sanitized.

I approve this to avoid blocking the merge if this fix is urgent.
I would prefer this contain a more general fix.

bjorkert:
- Renamed sanitizeToken → sanitizeConnectionInput (same whitespace/control-char stripping, now general for URLs and tokens).
  - processURL sanitizes the pasted value up front, fixing the URL-with-embedded-token + trailing whitespace/newline case that was hitting the lossy fallback and producing "Unknown Error".
  - Connect-time sanitizing now also covers the URL, not just the token.

marionbarker:
## Test

Using this build, insert a string that includes an extra space or return
* ✅ URL with embedded token handled correctly
* ✅ token handled correctly

marionbarker (approved):

PR #689, opened 2026-06-22
Skip alarm checks until BFU reload completes

bjorkert:
On a Before-First-Unlock cold launch, the alarm task can run before `performBFUReloadIfNeeded()` reloads StorageValues. During that window the alarm list (and its snooze state), `lastBGChecked`, and alarm configuration are still stale pre-unlock values, so a spurious Missed Reading alert can sound before real state is loaded.

`checkAlarms` now returns early while `needsBFUReload` is true; the next scheduled tick evaluates normally once the reload completes.

### DIFF
--- a/LoopFollow/Alarm/AlarmManager.swift
+++ b/LoopFollow/Alarm/AlarmManager.swift
@@ -42,6 +42,12 @@ class AlarmManager {
     }
 
     func checkAlarms(data: AlarmData) {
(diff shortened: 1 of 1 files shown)


bjorkert:
Superseded by #710 — the readiness gate keeps MainViewController (and the alarm task) from existing until storage is hydrated, so this guard is no longer needed.

PR #690, opened 2026-06-26
Parameterize bundle ID and build workflow for identical sister-repo code

bjorkert:
The sister apps (`LoopFollow_Second`, `LoopFollow_Third`) exist so users following 2–3 loopers can install separate apps with distinct bundle identifiers. They only need to differ from the main app by the bundle identifier and display name — both already driven by `app_suffix` / `display_name` in `LoopFollowDisplayNameConfig.xcconfig`. In practice a couple of build files still hardcoded `LoopFollow`, forcing them to diverge between the repos and making them a recurring merge-conflict source when changes are transferred on each release.

This makes those files identical across all three repositories and reworks how they are kept in sync:

**`fastlane/Fastfile`** — reads `app_suffix` from `LoopFollowDisplayNameConfig.xcconfig` and derives the app and Live Activity extension identifiers from it, replacing every hardcoded `com.<team>.LoopFollow` in the `build_LoopFollow`, `identifiers` and `certs` lanes. The same Fastfile now produces the correct identifiers everywhere:
- main: `com.<team>.LoopFollow` (unchanged)
- second: `com.<team>.LoopFollow.Second`
- third: `com.<team>.LoopFollow.Third`

Because the derived identifiers match what each repo is already provisioned for, existing certificates and provisioning profiles keep working unchanged.

**`.github/workflows/build_LoopFollow.yml`** — the not-configured summary now derives the variant name from `github.event.repository.name` / `UPSTREAM_REPO` instead of hardcoding it, and two comments no longer name the variant. The only intentional per-repo differences left are the scheduled `cron` minute and `UPSTREAM_REPO`.

**`release.sh`** — `update_follower` now mirrors the release tree into each sister repo with `rsync --delete` instead of generating and applying a patch, then re-applies that instance's overlay (`app_suffix`, `display_name`, `UPSTREAM_REPO`, `cron`) and renames the workspace. The per-instance values are derived from the sister directory name; only the staggered cron minute is passed in. Mirroring makes each sister an exact copy of the release tree except for the overlay, so transfers no longer hit patch-context conflicts and any accidental drift is corrected automatically. Only `.git`, local/build directories and the local team override are protected from the sync.

### DIFF
--- a/.github/workflows/build_LoopFollow.yml
+++ b/.github/workflows/build_LoopFollow.yml
@@ -131,10 +131,10 @@ jobs:
         if: steps.workflow-permission.outputs.has_permission != 'true'
         run: |
           echo "### :calendar: Scheduled Sync and Build Disabled :mobile_phone_off:" >> $GITHUB_STEP_SUMMARY
-          echo "You have not yet configured the scheduled sync and build for LoopFollow's browser build." >> $GITHUB_STEP_SUMMARY
-          echo "Synchronizing your fork of <code>LoopFollow</code> with the upstream repository <code>loopandlearn/LoopFollow</code> will be skipped." >> $GITHUB_STEP_SUMMARY
-          echo "If you want to enable automatic builds and updates for your LoopFollow, please follow the instructions \
-              under the following path <code>LoopFollow/fastlane/testflight.md</code>." >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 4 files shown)


marionbarker (approved):
Approve from code review.

marionbarker (approved):
I'm going to approve this from code review and merge it to dev.

I plan to test this prior to our next release using a set of 3 separate LF, LF_2 and LF_3 forks in an organization.  If doing a fake release using those throw-away forks reveals an issue with the release and mirroring process, then I'll come back with an issue for us to fix before we do the actual release.

PR #692, opened 2026-07-06
Add directional/charging skip toggles to Low, High and Battery alarms

bjorkert:
Adds three opt-in toggles that keep alarms quiet when the situation is already correcting itself. All are off by default, so existing alarms behave exactly as before.

- **Low BG — skip if rising:** stays silent while glucose is climbing, only firing when the latest reading is flat or still falling.
- **High BG — skip if falling:** the mirror case — stays silent while glucose is dropping, only firing when the latest reading is flat or still rising.
- **Phone Battery — skip while charging:** stays silent while the phone is plugged in, using the uploader's charging status from Nightscout. If the uploader doesn't report charging status, the alert still sounds.

For the BG alarms, direction is judged by the last two readings; the delta isn't configurable. The battery toggle relies on the `isCharging` flag already present in the Nightscout devicestatus payload rather than guessing from battery-level history.

### DIFF
--- a/LoopFollow/Alarm/Alarm.swift
+++ b/LoopFollow/Alarm/Alarm.swift
@@ -76,6 +76,10 @@ struct Alarm: Identifiable, Codable, Equatable {
     /// Number of minutes that must satisfy the alarm criteria
     var persistentMinutes: Int?
 
+    /// When set, the low alarm stays silent while BG is rising (positive delta),
+    /// only firing when the latest reading is flat or still falling.
+    var suppressIfRising: Bool = false
+
     /// Size of window to observe values, for example battery drop of x within this number of minutes,
     var monitoringWindow: Int?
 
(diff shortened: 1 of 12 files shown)


codebymini (approved):
👍 LGTM

marionbarker:
## Test

✅ successful test

**Battery**

Select skip when charging
Trio phone battery at 70%. Set battery alarm at or below 75%, get alarm
Plug in Trio phone. Change level to 90% to make sure battery stays below level
Battery shows 75% and plugged in - no alarm but wait a full 5 minutes to make sure.

**Low and High Alarms**

For the alarms - set persistent to 0 so they alert right away, set snooze to 0 minutes.
Create a cgm input file that goes up and down with 2 readings for each value.
Set low alarm at 90, high alarm at 180.
Should get only two alarms per event - once when cross threshold, once when flat.
pattern is 100, 100, 80, 80, 81, 82, 200, 200, 199, 198
upload at cadence of 150 sec, use a Trio instance to receive the values and update Nightscout URL

✅ alert at 80
✅ alert at 80
✅ silence at 81 until
✅ alert at 200
✅ alert at 200
✅ silence at 199

marionbarker (approved):
Approve from code review and test

PR #693, opened 2026-07-07
Add color thresholds to the Info Display

### LINKED ISSUE #691: New feaure idea: Info Display Color
Any thoughts on a new feature that would allow us to set ranges for yellow and red on the info display? It would allow a visual cue for things we need to pay attention to without triggering an alarm, to avoid alarm fatigue. See attached. Ignore BAGE; that's my own customization. 

Image

bjorkert:
Closes #691.

Info Display rows can now turn **yellow** or **red** when their value crosses a user-set threshold — a purely visual cue, no alarm. Supported rows: IOB, COB, battery, pump battery, TDD, rec bolus, carbs today, and sensor/cannula/insulin ages. The concerning direction is fixed per metric (battery colors when low, ages when high) and in-range values read green.

Under the hood the parallel `infoSort`/`infoVisible` arrays are replaced by a single ordered `[InfoDisplayItem]` store (one-time migration via `migrateStep10`, plus a safety-net conversion in `synchronizeInfoTypes`). The Info Display settings screen is reworked into drill-in rows that stay reorderable while also opening a per-row detail (visibility + colors).


marionbarker:
## Test

Once I was told how to enable this, it worked as expected. It should have been obvious but I missed seeing it in the Information Table settings.

Ready to merge once conflicts are resolved.

marionbarker (approved):
Approve by code review and test. (Test repeated after last push).

PR #694, opened 2026-07-07
Add search to the alarms list

bjorkert:
Adds a search field to the alarms list so alarms can be filtered while browsing. Matching is case-insensitive and covers both the customized alarm name and the underlying alarm type name.

Uses the same `.searchable` presentation and "No Results" empty state as the Add Alarm sheet's search.


marionbarker (approved):

PR #696, opened 2026-07-08
Add dismissable top banner for Nightscout, Dexcom, and heartbeat issues

bjorkert:
## Summary

Failed BG downloads are currently only visible in the log viewer, and a failed Dexcom Share login silently falls back to Nightscout — a user with a bad token, mistyped URL, or wrong Dexcom credentials sees no data with no explanation.

This adds a dismissable banner above the tab view. It is visible on every tab and pushes the content down (up to 3 lines) rather than overlaying it.

## What it does

- **Nightscout failures**: after a failed BG fetch, a rate-limited probe of `status.json` classifies the cause, so the banner can say *why* — "The token is invalid.", "A token is required.", "The site was not found.", "A network error occurred." — reusing the same `NightscoutError` messages the settings screen shows. If the server is reachable but the data fetch still failed, a softer warning says so.
- **Dexcom Share failures**: `ShareError` login codes are mapped to friendly text (account not found, incorrect username or password, too many failed attempts). Matching is by substring so both the legacy `SSO_Authenticate…` codes and the current `AccountPasswordInvalid` are covered. When Nightscout is configured as backup, the previously silent fallback is now surfaced as a warning ("… Using Nightscout as backup.") while data keeps flowing.
- **Frequent heartbeat dropouts**: when the Bluetooth heartbeat becomes regularly delayed — 5 or more beats arriving later than 1.5× the expected interval within the last hour, a typical symptom of a dying transmitter battery — a warning suggests checking the transmitter/RileyLink/pod battery or Bluetooth range. A single late beat never triggers it; gaps large enough to indicate out-of-range, Bluetooth off, or a suspended app reset the count instead of adding to it. The warning clears after a clean hour, or immediately on device disconnect/switch.
- **Clearing**: a banner disappears automatically when its source recovers (next successful fetch, clean heartbeat window) or when that source is deconfigured (URL/credentials removed, device disconnected).
- **Dismissal**: tapping ✕ hides the banner. The same error stays hidden for 30 minutes (then re-appears if still occurring); a *different* error shows immediately. Repeated failing retries reuse the same message, so the banner doesn't re-animate every 10–60 s cycle.
- **Severity**: when multiple sources have problems, the more severe message wins (error > warning > info); dismissing it reveals the next one.

The banner system is generic (`BannerManager.report/clear` with a `custom` source), so other important notices can use it later.

## Also fixed

`NightscoutUtils.verifyURLAndToken` classified transport errors backwards: DNS/connection failures were reported as `siteNotFound` and everything else as `networkError`. It now distinguishes them by `URLError` code, which also improves the status label in Nightscout settings.


bjorkert:
image

marionbarker (approved):
Approved by code review and minimal test.
It builds.
If I mess up one item (in this case Nightscout token), I get a dismissable banner.
Once dismissed, it returns following a quit restart.

PR #697, opened 2026-07-08
Add Nightscout database size info row and alarm

bjorkert:
Nightscout's `dbsize` plugin exposes the database size at `/api/v2/properties/dbsize`. A new scheduler task polls it every 6 hours and surfaces the result in two places.

**Info row** — a new optional `DB Size` row showing used MiB and the percentage of the site's configured limit. Hidden by default.

**Alarm** — a new `Nightscout Database Size` alarm firing at or above a percentage threshold, defaulting to 75% to match the plugin's own urgent level. The database fills over weeks, so it defaults to daytime only.

The percentage is measured against `DBSIZE_MAX`, which is 496 MiB unless the site owner changed it. The info row therefore leads with the absolute MiB figure, and the alarm editor explains the caveat.

### DIFF
--- a/LoopFollow.xcodeproj/project.pbxproj
+++ b/LoopFollow.xcodeproj/project.pbxproj
@@ -7,10 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		B500000000000000000000A2 /* RemoteBolusHistoryEntry.swift in Sources */ = {isa = PBXBuildFile; fileRef = B500000000000000000000A1 /* RemoteBolusHistoryEntry.swift */; };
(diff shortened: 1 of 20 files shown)


bjorkert:
CleanShot 2026-07-08 at 21 39 51@2x
CleanShot 2026-07-08 at 21 41 13@2x

codebymini (approved):
👍

marionbarker:
## Test

✅ successful test

The most full Nightscout site I have is only 3% full. Modify the step size from 5% to 2% so I can set 2% as the alarm threshold.

🤣 Love the alarm.

marionbarker (approved):

PR #700, opened 2026-07-08
Format Rec. Bolus with the pump's bolus increment

bjorkert:
Rec. Bolus was formatted two different ways. The Loop device status parser used a hardcoded `String(format: "%.2fU")`, while the OpenAPS/Trio parser went through `InsulinMetric`, which drops to one fraction digit below 10 U. The same recommendation rendered as `0.00U` in the info table on a Loop URL and `0` on a Trio URL.

Both parsers now use `InsulinFormatter`, which derives its fraction digits from the pump's reported `bolusIncrement`, so Rec. Bolus is shown at the precision the pump can actually deliver — two decimals for a 0.05 U DASH pod. The unit suffix is dropped to match the other insulin rows in the info table.

When the device status carries no recommendation, the info row and `deviceRecBolus` are now cleared. Previously the Loop parser left both stale, so an old recommendation could linger in the info table, the Rec. Bolus alarm condition, and the Live Activity.

### DIFF
--- a/LoopFollow/Controllers/Nightscout/DeviceStatusLoop.swift
+++ b/LoopFollow/Controllers/Nightscout/DeviceStatusLoop.swift
@@ -100,10 +100,9 @@ extension MainViewController {
                 infoManager.clearInfoData(type: .minMax)
                 updatePredictionGraph()
             }
-            if let recBolus = lastLoopRecord["recommendedBolus"] as? Double {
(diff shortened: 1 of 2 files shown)


codebymini (approved):
👍 Approve by code review and test with DASH pod connected to Loop

marionbarker (approved):

PR #702, opened 2026-07-10
Migrate main LoopFollow target to a synchronized folder group

bjorkert:
Converts the main `LoopFollow/` group to a `PBXFileSystemSynchronizedRootGroup`, so files are synced from disk instead of enumerated by hand. `project.pbxproj` drops from ~2900 to ~880 lines and stops generating merge conflicts when files are added, moved, or removed.

- Sources shared with the Live Activity extension move to a new `Shared/` folder referenced by both targets.
- `Info.plist` is excluded from the copy phase via a membership exception.
- Removes dead code that was on disk but never built: a stale duplicate `RestartLiveActivityIntent`, four unreferenced orphan files, and a Core Data model unused since the initial commit.

Builds clean via the workspace (app + embedded extension).


marionbarker (approved):
Approve by test. Builds to iOS 18, iOS 26.
The Mac Catalyst build was fixed with PR #715. Apply that fix to this PR and builds to Mac Catalyst.

PR #703, opened 2026-07-12
Migrate charts to Swift Charts

bjorkert:
Replaces the danielgindi Charts and SwiftAlgorithms CocoaPods with native Swift Charts.

- Rebuilds the BG graph, treatments, and prediction cone on Swift Charts with a custom pan/zoom/inspect interaction layer.
- Ports the stats views (AGP, TIR, GRI) off the old Charts dependency.
- Drops the Charts and SwiftAlgorithms pods and their generated support files.

Chart appearance and behaviour are kept at parity with the previous implementation.


bjorkert:
Superseded by #704 (same change, single commit).

PR #704, opened 2026-07-12
Migrate charts to Swift Charts

bjorkert:
Replaces the danielgindi Charts and SwiftAlgorithms CocoaPods with native Swift Charts.

- Rebuilds the BG graph, treatments, and prediction cone on Swift Charts with a custom pan/zoom/inspect interaction layer.
- Ports the stats views (AGP, TIR, GRI) off the old Charts dependency.
- Drops the Charts and SwiftAlgorithms pods and their generated support files.

Chart appearance and behaviour are kept at parity with the previous implementation.


dnzxy:
Hyped to see this land 🚀

codebymini:
The work in `fix/forecast-clear-on-device-switch` needs to be recreated here since the Graphs.swift file is removed

codebymini (approved):
👍 LGTM

marionbarker (approved):
Approve by test and code review by the experts.
Built on test phone and real phone.
Viewed Trio URL and Loop URL.

PR #707, opened 2026-07-16
Only clear the forecast when the data source changes

### LINKED ISSUE #543: Prediction missing on initial Loop URL render until next CGM update
When it first renders a Loop URL, the plot is slow to appear and the prediction is not updated until the next CGM reading.

bjorkert:
The refresh handler cleared the Loop and Trio forecasts on every refresh, including every websocket disconnect. On Loop that made the forecast vanish and only reappear on the next loop cycle, so on some instances it looked like it never came back.

The clearing now only happens when the active system actually changes. When a device status flips between Loop and Trio the previous system's forecast is dropped, and when Nightscout is turned off so there is no device status source, such as Dexcom only, both forecasts are cleared. Normal refreshes and websocket reconnects no longer touch the forecast.

Also rebuilds the Loop forecast whenever it is empty, not only when the loop cycle advances, so it is restored on the next device-status poll regardless of cycle timing.

For pumps that report no `pump.clock`, the loop time now falls back to the loop cycle timestamp. Without it the loop time stayed 0, so the forecast anchored to epoch 0 and stretched the graph across decades.

### DIFF
--- a/LoopFollow/Controllers/Graphs.swift
+++ b/LoopFollow/Controllers/Graphs.swift
@@ -967,6 +967,30 @@ extension MainViewController {
         BGChart.notifyDataSetChanged()
     }
 
+    // Removes the Loop forecast line. Used when the active system switches away from Loop.
+    func clearLoopPredictionGraph() {
+        guard !predictionData.isEmpty else { return }
+        predictionData.removeAll()
+        updatePredictionGraph()
+    }
+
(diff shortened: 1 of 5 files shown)


codebymini (dismissed):
👍 

marionbarker (approved):
Approve from code review and test with Loop URL from next-dev and Trio URL

Thanks

PR #709, opened 2026-07-18
Add fallbacks for missing TDD and timestamp in devicestatus

bjorkert:
Some Trio uploads have a suggested object without TDD and timestamp, which left the TDD and Updated rows blank until the next loop cycle.

TDD now falls back to the enacted object when suggested doesn't have it, and Updated uses deliverAt with timestamp as fallback.


marionbarker (approved):
Approve from code review.

PR #710, opened 2026-07-18
Gate startup on storage readiness for Before-First-Unlock recovery

bjorkert:
**What**
Gates app startup on storage readiness. Before a device's first unlock after a reboot (Before-First-Unlock / BFU), UserDefaults is encrypted and every stored value reads as its default. This detects that state via a presence sentinel, holds the UI behind a lightweight loading view until the store is readable, then hydrates every value and reconnects Bluetooth. Replaces the hand-maintained `reloadAll()` recovery — values now register for hydration automatically, so none can be missed.

**Why**
On a BFU background launch (BLE or background-refresh wake) the app otherwise acts on poisoned defaults: it re-prompts telemetry consent, runs with an empty Nightscout URL, and leaves the CGM heartbeat down after recovery. Ordinary locked launches are unaffected.

**How to test**
Requires a real device with a passcode (the Simulator can't reproduce BFU). Configure a data source, reboot the phone and — *without unlocking* — let a BLE/background wake fire. Then unlock and open the app; the logs should show `BFU check:` and `BFU recovery complete: url='…'`. Expect no telemetry re-prompt, Bluetooth reconnected, and a non-empty url.

**Note**
The BFU sentinel ORs a few presence probes (`migrationStep`, `telemetryConsentDecisionMade`, `url`) alongside a dedicated marker, to correctly classify users who haven't launched this build yet. About a year after release — once all users are expected to have the marker — those extra probes can be dropped in favour of the marker alone.


marionbarker:
## Test

Build onto SE 3rd gen, iOS 26 device.
I already have an EmaLink configured as background refresh source.

Reboot the phone but do not unlock it right away
11:02 Phone reboot is complete
11:06 unlock phone and check log - BFU recovery is noted. There was also an alert about it.

marionbarker (approved):

PR #711, opened 2026-07-19
Remove pre-iOS 18 Live Activity code paths

bjorkert:
Now that the deployment target is iOS 18, this removes the Live Activity code paths and availability guards that only ever ran on older systems.

### Dead code removed
- **`LiveActivityManager`** — deleted `startIfNeededLegacy` and `attemptLegacyRenewal` (the `Activity.request()`-based start and renewal fallbacks), unwrapped the four `#available(iOS 17.2, *)` push-to-start guards, and dropped the three now-redundant `@available(iOS 17.2, *)` attributes. Push-to-start is the single creation path.
- **`LoopFollowLiveActivity`** — collapsed the dual `ActivityConfiguration` into the `.small`-family-adaptive layout, inlined `.contentMargins(.all, 0)` (removing the `applyActivityContentMarginsFixIfAvailable` shim), and removed the `iOSApplicationExtension 16.1` / `iOS 18.0` attributes.
- **`StorageCurrentGlucoseStateProvider`** — `showRenewalOverlay` now returns `laRenewalFailed` directly.
- **`EKEventStore+Extensions`** — requests full calendar access directly instead of branching on iOS 17.
- **`RestartLiveActivityIntent`** — dropped the redundant `@available(iOS 16.4, *)` attributes (the `iOS 26.0` gate for `continueInForeground()` is kept).

### Also
- Deleted the orphaned `LoopFollow/LiveActivity/RestartLiveActivityIntent.swift`, which was not referenced by any target and shadowed the maintained root-level copy.


marionbarker (approved):
Approve from code review.
Test that it builds.
Test that LA works for iOS 18 and iOS 26 test phones.

PR #712, opened 2026-07-19
Use NightscoutUtils.constructURL for the Nightscout web viewer URL

bjorkert:
The Nightscout web viewer built its URL by concatenating the token onto the base URL without any encoding, so a token containing characters like & or = would be silently mangled. It now goes through NightscoutUtils.constructURL, which percent-encodes the token. For a normal token the resulting URL is identical to before.

constructURL also skips assigning queryItems when there is nothing to add, so a setup without a token no longer gets a stray trailing question mark.


codebymini (approved):

marionbarker (approved):
approve from code review and test

PR #713, opened 2026-07-19
Remove unneeded UIKit imports

bjorkert:
Removes `import UIKit` from 16 files that don't use any UIKit API. `BackgroundRefreshManager.swift` gets `import Foundation` instead, since it was relying on UIKit re-exporting Foundation. Verified with a full workspace build.


marionbarker (approved):
- Confirm this builds to iOS 26 iPhone
- Discover Mac Catalyst is broken - but not from this PR (see #715)

PR #714, opened 2026-07-19
Add new section to export APNS setting

codebymini:
APNS import/export disappeared when APNS moved to its own settings section.
This PR restores parity so users can transfer APNS credentials via QR just like other configuration categories.

### User Impact
Users can now export APNS settings to QR.
Users can import APNS settings from QR.
Users get clearer warning messaging before overwriting APNS credentials.
Export/import copy is more consistent and readable.


marionbarker (approved):
approved by code review and test.

PR #695, opened 2026-07-08
Add search to the Menu

bjorkert:
Adds a search field to the **Menu** tab, matching the iOS 26 style of the in-flight search additions to the Add Alarm sheet (#681) and the alarms list (#694).

Search covers the Menu's own rows (Settings, the tab features, View Log / Share Logs, and the support links) **and** the rows inside the Settings sub-screen (General, Graph, Units, Alarms, …). Typing e.g. "graph" surfaces the Graph settings row and tapping it deep-links straight into that screen.

### Single source of truth
To avoid maintaining a parallel search index, the menu is now driven from the data that also feeds the search:

- `SettingsRoute` carries each row's `title`, `icon` and search `keywords`, and a new `SettingsSection` + `menuSections(nightscoutConfigured:)` groups the rows and encodes their conditional visibility (Information Display / Remote require Nightscout; Live Activity is excluded on Mac Catalyst) in one place. `SettingsMenuView` renders from this — visually unchanged.
- The Menu search results are assembled from those same routes, `TabItem.featureOrder`, and the handful of static rows, so adding a settings page automatically appears in both the list and the search with nothing to keep in sync.

Empty search shows the normal menu; a query with no matches shows a "No Results" state consistent with #681/#694.

PR #715, opened 2026-07-19
Fix Mac Catalyst build

bjorkert:
The Menu search keywords added in #695 referenced the liveActivity settings route without the macCatalyst guard the rest of the code uses, which broke the Mac Catalyst build. This wraps that case in the same conditional. Both the Catalyst and iOS builds succeed.


marionbarker (approved):

PR #717, opened 2026-07-21
Fix graph settings not honored on the small graph

bjorkert:
The Swift Charts migration (#704) lost some of the settings that control
what the small graph shows.

- "Treatments on Small Graph" was ignored. Treatments, basal, overrides
  and temp targets were always drawn, and flipping the toggle didn't
  refresh the chart.
- Midnight lines no longer showed on the small graph.
- With prediction style "Lines" the ZT/IOB/COB/UAM lines were missing
  from the small graph, so Trio users had no prediction there.
- Yesterday's BG was drawn on the small graph. The old chart only drew
  it on the main graph.

This restores the old behavior. The prediction cone stays visible on the
small graph. Also added the missing chart refresh to the "Show Carb/Bolus
Values" and "Show Carb Absorption" toggles so they apply right away
instead of on the next reading.

Tested in simulator:
- Toggling "Treatments on Small Graph" off hides boluses, carbs, SMBs,
  basal, overrides and temp targets on the small graph only, and they
  come back when toggled on again, without restarting the app.
- Midnight lines show on both graphs when enabled.
- Yesterday's BG shows on the main graph only.
- Cone style shows the cone and Lines style shows all four prediction
  lines on both graphs.
- Built for device and simulator via the workspace.

### DIFF
--- a/LoopFollow/Charts/BGChartModel.swift
+++ b/LoopFollow/Charts/BGChartModel.swift
@@ -174,6 +174,7 @@ final class BGChartModel: ObservableObject {
     @Published var show30Min: Bool = false
     @Published var show90Min: Bool = false
     @Published var showMidnight: Bool = false
+    @Published var smallGraphTreatments: Bool = true
 
     private static let doseFormatter: NumberFormatter = {
         let nf = NumberFormatter()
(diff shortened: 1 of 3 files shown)


marionbarker:
## Test

✅ Trio URL (on SE 3rd gen running iOS 26)
✅ Loop URL (on SE 2nd gen running iOS 18)
✅ Loop URL (on Mac Catalyst)

marionbarker (approved):

PR #718, opened 2026-07-21
Move main chart in realtime while dragging on the small graph

bjorkert:
Dragging on the small graph now moves the main chart to the finger position live while dragging, instead of only jumping there on release. A plain tap still behaves like before and animates to the tapped position.


marionbarker:
## Test

✅ Trio URL (on SE 3rd gen running iOS 26)
✅ Loop URL (on SE 2nd gen running iOS 18)
✅ Loop URL (on Mac Catalyst)

marionbarker (approved):

PR #719, opened 2026-07-23
Show active override and temp target with remaining time on the Live Activity and in the remote command views

bjorkert:
The Live Activity gets a new row between the metric grid and the footer that shows the active override and/or temp target. The row only appears while one of them is active. When 2 hours or less remain, a live countdown ticks next to the name or target value. Longer or indefinite overrides show just the name, since a multi-day ticker gets silly. The same row appears in the expanded Dynamic Island above the updated-at line.

Since the row makes the Override grid slot redundant, it is removed from the four grid slot pickers. Stored selections from older versions render as an empty cell. It is still selectable for the CarPlay/Watch slot, which has no row.

The remote command views also show the remaining time of the active override or temp target: the Trio override and temp target views get a Time Remaining row, and the Loop override presets view shows it under the active override name. A countdown when 2 hours or less remain, otherwise a static duration like 5h 30m. Durations of a week or longer are treated as indefinite and show no remaining time at all, since Trio represents indefinite overrides and temp targets as 30 day durations.

On the remote control overview, the Overrides and Temp Target buttons get a green ring and glow while the corresponding override or temp target is active.

The end times come from the treatments parsing and travel to the extension via three new optional snapshot fields. Indefinite Loop overrides ship no end time, and the graph-edge clamp used for the chart bands is not applied to these, so long overrides count down correctly. Also clears the active override and temp target observables when treatments age out of the download window, which previously left them stale.

### DIFF
--- a/LoopFollow/Controllers/NightScout.swift
+++ b/LoopFollow/Controllers/NightScout.swift
@@ -72,11 +72,15 @@ extension MainViewController {
 
     func clearOldOverride() {
         overrideGraphData.removeAll()
+        Observable.shared.override.value = nil
+        Observable.shared.overrideEndAt.value = nil
         updateOverrideGraph()
     }
 
     func clearOldTempTarget() {
         tempTargetGraphData.removeAll()
+        Observable.shared.tempTarget.value = nil
+        Observable.shared.tempTargetEndAt.value = nil
         updateTempTargetGraph()
     }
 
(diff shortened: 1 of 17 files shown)


bjorkert:
FE585A76-C271-4C20-AEB6-885A6B0C23E3_1_105_c
B7E4726E-E044-4B7F-8207-6547E9E6AF32
22B6C60B-2F61-40A0-B617-B9C54F22EEA6

marionbarker:
## Test

✅ successful test

### Test with Trio URL

✅ Confirm the extra row seen for override and temp target on Live Activity when they are active.

### Test with Loop URL

✅ Confirm the extra row seen for preset on Live Activity when it is active.

### Test Mac Catalyst

✅ Confirm build works on mac (Live Activity code is protected by targetEnvironment check - not available on Mac)

marionbarker (approved):
Approved by code review and test

PR #720, opened 2026-07-23
Draw temp targets at the BG level they target instead of in the override strip

bjorkert:
The Swift Charts migration put temp target bands in the same strip at the top of the graph as overrides, so the two collided and the band no longer showed which BG level the temp target aims for. Before the migration the band was drawn at the target value, 5 mg/dL above and below it, with no label, and only overrides used the top strip. This restores that behavior. Verified with an override and a temp target active at the same time, on both the main and the small graph.

### DIFF
--- a/LoopFollow/Charts/BGChartModel.swift
+++ b/LoopFollow/Charts/BGChartModel.swift
@@ -537,11 +537,14 @@ final class BGChartModel: ObservableObject {
         }
         tempTargets = vc.tempTargetGraphData.map {
             let target = $0.correctionRange.first.map { String($0) } ?? ""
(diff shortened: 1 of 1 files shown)


marionbarker:
## Test

✅ successful test

### Reproduce the problem

start an override and temp target on the Trio test phone and confirm they are overlapped in the LoopFollow display with dev branch

### Confirm the fix

Build the new version of LoopFollow and confirm the Temp Target is plotted at appropriate glucose levels.
The Override is labeled and in a bar above the glucose range as expected.

> `Temp Target is plotted at` target value, 5 mg/dL above and below it, with no label, and only overrides used the top strip.

marionbarker (approved):

PR #721, opened 2026-07-23
Add an early warning option to the temp target and override end alarms

bjorkert:
This adds an Early Warning setting to the Temp Target Ended and Override Ended alarms. It can be set from 0 to 30 minutes in steps of 5 and defaults to 0, which keeps the current behavior. With a lead time set, the alarm fires that many minutes before the scheduled end with the title Temp Target Ending Soon or Override Ending Soon, and then fires again as usual when it actually ends. Each firing happens at most once per event, and extending an active temp target or override arms a new warning for the new end time.

Overrides now carry their unclamped scheduled end next to the graph end date, which is capped to the display horizon. Indefinite Loop overrides have no known end and never trigger the early warning. The full screen alarm view shows the early warning title too.

Includes unit tests for both alarm conditions.


bjorkert:
image

PR #723, opened 2026-07-24
Use Trio's colors for the forecast curves

bjorkert:
The Swift Charts migration hardcoded system colors for the Trio/OpenAPS prediction lines, so they no longer matched what Trio shows. This switches ZT, IOB and UAM back to the ZT, Insulin and UAM asset colors, which are identical to the ones Trio uses. COB already renders in the same orange as Trio, so it is unchanged.


marionbarker (approved):

* Built on Mac Catalyst (Loop URL)
* Built on SE 2nd gen (Loop URL)
* Built on SE 3rd gen (Trio URL)
   * observe prediction colors are updated to match Trio colors with this PR

PR #727, opened 2026-07-25
Add App Intents to enable/disable BG speech via Shortcuts

skimaniac:
## What
Adds two App Intents (Enable/Disable BG Speech) so users can toggle
speakBG from the Shortcuts app or from a personal automation.

## Why
speakBG is useful when wearing Bluetooth headphones, but there's
currently no way to enable/disable it automatically (e.g. when
headphones connect/disconnect) without opening the app. This exposes
the existing speakBG toggle as two Shortcuts actions so users can
build their own automations (Bluetooth-triggered, time-based, etc.).

## Testing
Built and tested on-device (iOS 26.5.2) against dev branch. Confirmed
both actions appear in the Shortcuts app and correctly toggle
Storage.shared.speakBG.


bjorkert (approved):
Thanks for the contribution!

PR #729, opened 2026-07-26
Fix Advanced Settings graph toggles not hiding chart elements

### LINKED ISSUE #728: Advanced Graph display options broken
7.0 install
iPhone Air
IOS 26.5.2

The advanced setting toggles to hide the different treatments on the graph no longer work to hide them.  My daughter is not looping, but I pull her Omnipod carbs/bolus into NS via Glooko. So I leave on all of those except basal since that has no valid data coming in. The new graph shows the dashed blue lines from a basal profile saved in NS even those graph basal is disabled.

bjorkert:
The Graph Basal, Graph Bolus, Graph Carbs and Graph Other Treatments toggles in Advanced Settings stopped working in 7.0. The old chart only drew what the gated update calls pushed into it, but the new chart model reads the data arrays directly, so the toggles were never consulted. The model now skips the corresponding data when a toggle is off, and flipping a toggle marks the chart dirty so the graph redraws when you return to the home screen.

Two more settings-change behaviors were lost in the migration and are restored here: changing the prediction style re-routes the stored predBGs again, so switching between cone and lines no longer waits for the next device status fetch, and toggling Show Yesterday's BG reloads the BG window so the overlay appears or disappears right away instead of at the next scheduled fetch.

Verified in the simulator against a live Nightscout site: each toggle hides and restores its elements on both the main and the small graph, and the yesterday line appears within seconds of enabling it.

Fixes #728

### DIFF
--- a/LoopFollow/Charts/BGChartModel.swift
+++ b/LoopFollow/Charts/BGChartModel.swift
@@ -408,6 +408,14 @@ final class BGChartModel: ObservableObject {
         showMidnight = Storage.shared.showMidnightLines.value
         smallGraphTreatments = Storage.shared.smallGraphTreatments.value
 
+        // Advanced-settings visibility toggles. The Nightscout controllers
+        // collect the data regardless (it also feeds the info rows), so hidden
+        // kinds are dropped here at render time.
+        let showBasal = Storage.shared.graphBasal.value
+        let showBolus = Storage.shared.graphBolus.value
+        let showCarbs = Storage.shared.graphCarbs.value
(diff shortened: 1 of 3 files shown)


codebymini (approved):
👍 LGTM Tested with Loop Data

marionbarker (approved):
Approve by testing the build:

* ✅ macCatalyst, Loop URL
* ✅ iOS 26 phone, Trio URL
* ✅ iOS 18 phone, Loop URL

enable/disable buttons work as expected.

PR #733, opened 2026-07-29
Fix main graph locking up after system-cancelled gestures

bjorkert:
A few times the main graph has stopped responding to panning and zooming while the small graph could still move it, and only an app restart would unlock it. The cause is that SwiftUI gestures never receive onEnded when the system cancels a touch, for example from a permission alert, an incoming call or an app switch. The pinch or inspect state then stays latched and swallows every following touch, and in the worst case the gesture system for the view stops delivering events entirely.

This change clears leaked gesture state whenever a new touch begins, resets it when the chart disappears, and rebuilds the chart's gesture attachments when the app returns to the foreground. The current position and zoom are kept across that rebuild, so nothing changes visually.

### DIFF
--- a/LoopFollow/Charts/BGChartModel.swift
+++ b/LoopFollow/Charts/BGChartModel.swift
@@ -17,6 +17,10 @@ final class BGChartInteraction: ObservableObject {
     /// when the user pans back into history, re-armed when they return to the edge.
     @Published var followLatest: Bool = true
 
+    /// Set once the main chart has done its initial scroll-to-now; kept out of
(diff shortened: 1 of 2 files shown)


marionbarker (approved):
Accepting code from @bjorkert .

Test by building and using the main-screen zoom and scroll features.
* ✅ Mac Catalyst (not the pinch part)
* ✅ SE 2nd gen, iOS 18, Loop URL
* ✅ SE 3rd gen, iOS 26, Trio URL

PR #739, opened 2026-08-14
Bugfix hover repeat entries

aug0211:
Without this fix, SMBs/boluses/etc can be duplicated when sliding the finger along the graph - making it look as if multiple SMBs/etc were sent when in fact only 1 was. This fixes that behavior so that each entry on the graph lives as its own distinct point when sliding across the graph.


bjorkert:
Thanks for pointing this out, https://github.com/loopandlearn/LoopFollow/pull/748 replaces this PR.

PR #740, opened 2026-08-14
Refactor OverrideActivationModal to simplify duration handling

### LINKED ISSUE #730: Edit Loop Remote Override Duration
LoopFollow does not allow the user to manually define the duration of a remote override if that override has a preset duration defined. 

Yet, you can change the duration for indefinite remote overrides. 

Can we add a way to edit duration for all overrides? Maybe reuse the existing UI elements that make the indefinite toggle visible and a slider when the indefinite toggle is off.

One use case: Looper forgets to put in sports override that lasts for the duration of practice plus 1 hour (total 2.5 hours). I notice before practice starts and put on the override. Now, the override is on for at least 1 hour after practice, which I don't want, and now I need to remember to turn it off.

Second use case: dealing with a high or long meal and I want to run a low target override for longer than the defined 1 hour (which is set for safety).

# **Replication**
## Indefinite overrides (desired behavior)

Remote > override > select override that has no duration (indefinite=yes). Green toggle appears in the bottom right for indefinite duration. Toggle this off and a slider appears to set duration. 

## Override with duration. Undesired behavior.
Remote > override > select override that has a defined duration (not indefinite). There is not option to change duration.

Thanks, all.

codebymini:
This PR removes the timing restriction for remote override activation and allows users to set a custom duration for every override preset, including presets that already have a predefined duration.

What changed
Enabled duration controls for all override presets in the activation modal.
Kept sensible defaults:
Presets with fixed durations default to their preset duration.
Indefinite presets default to “Enable indefinitely”.
Updated activation behavior so the submitted duration always follows user choice:
“Enable indefinitely” ON sends an indefinite override.
“Enable indefinitely” OFF sends the selected slider duration.
Why
Previously, presets with predefined durations could not be adjusted at activation time. This prevented quick context-based adjustments. The new behavior gives users consistent, flexible control without losing preset defaults.

UX impact
More consistent activation flow across all presets.
Faster adjustments for real-world situations.
Backward-friendly behavior through preset-based defaults.

fixes https://github.com/loopandlearn/LoopFollow/issues/730


bjorkert (approved):
LGTM

marionbarker (approved):
* tested Loop enables Override duration to be change regardless of saved value on Loop phone
* tested Trio is not modified

PR #741, opened 2026-08-16
Fix missing Trio FPU entries

aug0211:
- LoopFollow was deduplicating Nightscout treatments by Trio's shared FPU ID, which hid the first scheduled entry from the graph
- Deduplicate by event type and occurrence time as well, preserving distinct FPU entries while still collapsing true duplicates

### DIFF
--- a/LoopFollow/Controllers/Nightscout/Treatments.swift
+++ b/LoopFollow/Controllers/Nightscout/Treatments.swift
@@ -4,6 +4,39 @@
 import Foundation
 
 extension MainViewController {
+    private struct TreatmentOccurrence: Hashable {
+        let id: String
+        let eventType: String?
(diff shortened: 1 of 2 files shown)


bjorkert (approved):
Verified against Trio's uploader: FPU siblings are uploaded with the shared fpuID as `id` and only `created_at` (Trio's NightscoutTreatment has no `timestamp` field), with entries spaced by the minute interval (min 10), so the occurrence key always splits siblings correctly. Ran the suite plus a simulator A/B against mock Nightscout data: dev collapsed 5 siblings to one dot (carbs today 32), with this fix all 5 render (40) and true duplicates (same id, same created_at, different _id) still collapse — the #569 behavior is preserved.

PR #745, opened 2026-08-21
Recover the Silent Tune keep-alive after audio interruptions

bjorkert:
Silent Tune's looping silent audio is the app's only background claim. When it stops, iOS suspends the app within seconds and it cannot recover on its own, which is what produces the "App inactive for N minutes" notification.

Reactivation now runs inside a UIApplication background task assertion so the retries get runtime even without the audio claim, keeps trying for about 18 seconds, and no longer trusts player.isPlaying, which keeps reporting true after the session is taken. The BGAppRefreshTask stays open until the restart resolves, and scheduleRefresh no longer pushes a pending request further out.

Three other things stop the audio and none were handled. A route disappearing, such as leaving CarPlay, pauses the player with no interruption notification. A media services reset invalidates the session and player. And a process launched into the background by BGAppRefreshTask never runs the backgrounding transition, so it had no observers attached and no background alerts armed. All three are covered now, and a background recovery clears any delivered notification it has made obsolete.

This does not make Silent Tune reliable, since an interruption that outlasts the assertion still suspends the app. Logging is quiet while healthy and detailed once struggling, including one line per lost runtime window measured against a monotonic clock.

### DIFF
--- a/LoopFollow/Controllers/BackgroundAlertManager.swift
+++ b/LoopFollow/Controllers/BackgroundAlertManager.swift
@@ -64,14 +64,14 @@ class BackgroundAlertManager {
     func scheduleBackgroundAlert(force: Bool = false) {
         guard isAlertScheduled, Storage.shared.backgroundRefreshType.value != .none else { return }
 
-        // Throttle execution if not forced: only run once every 10 seconds.
-        if !force {
-            let now = Date()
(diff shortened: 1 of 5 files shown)


marionbarker (approved):
Approve by test build and DM from others using this version.

PR #746, opened 2026-09-07
Carry the pump reservoir value across records that omit it

bjorkert:
LoopFollow reads the newest Nightscout device status record and treats a missing pump.reservoir as a reservoir above 50U. That is right for a full Omnipod, but Loop also leaves the field out of many records while the pod is below 50U.

I looked at 13 days of data from two Nightscout sites. There were 28 and 40 such gaps, median 25 minutes, the longest 85 minutes, and several of them happened while the pod held between 7 and 19 units. The pump row read 50+U through all of them, and since latestPumpVolume was pinned to 50 the low reservoir alarm could not fire either.

LoopFollow now keeps the last exact reading per pump and reuses it for 30 minutes when the field is missing. Once the reading is older than that the row shows an em dash. Replaying the same 13 days, the 172 and 228 records that read 50+U while the pod was below 50U drop to none, 127 and 157 records show the real volume, and 63 and 88 show the em dash.

Loop's first records for a new pod still carry the previous pod's final volume for up to 10 minutes, so a reading from the first 15 minutes after a pump appears is not taken as evidence that the pod is below 50U.

Trio and iAPS put no pumpID, manufacturer or model in the record, and they upload the reservoir without gaps, so records with no pump identity are resolved from the record alone. I checked six Trio and iAPS sites and no record resolves differently there. A Medtrum pump on Trio reports its volume in every record and is not affected.

Pumps that do name themselves and report a volume in every record show an em dash when the field is missing, since 50+U only makes sense for a pump that reports a volume solely when it is low.

Docs for the pump row will need a note about the em dash, in a separate PR.


codebymini (approved):
Approve by code review and test

marionbarker:
## Test

✅ successful test

I accidentally hit a Loop situation where the Nightscout site is not showing my current reservoir value but my LoopFollow phone is.

### Configuration

Built `fix/pump-reservoir-dropouts` to my personal phone several days ago
* this phone has both Loop and LoopFollow on it

### Test Narrative

Check LoopFollow Information display
* Pump reservoir is currently at 26 U
* On Loop phone, reservoir is 26.3 U
   * basal rate is 0.65 U/hr 
* Check the Nightscout site - it says pump reservoir is 50+ U, which is not correct
* ✅ the fact that LoopFollow shows correct amount is a successful test

marionbarker (approved):
Approve by test.

PR #747, opened 2026-09-15
Zip log files before sharing

bjorkert:
Log files can get large enough that mail apps warn about the attachment size when sharing them. Share Logs now puts the notice file and the log files into a single zip archive and hands that to the share sheet. The archive is named LoopFollow Logs YYYY-MM-DD_HHMM.zip and unpacks to a folder with the same name.

The zip is created with NSFileCoordinator and the forUploading option, so there is no new dependency. Zipping runs off the main thread. If it fails, the error is logged and the files are shared uncompressed as before.


codebymini (approved):
Approve by code review

marionbarker (approved):
Approve by test.

PR #748, opened 2026-09-17
Snap chart scrubbing to a five-minute grid

bjorkert:
While scrubbing the main chart, the indicator now moves in five-minute steps instead of stopping on every treatment. The grid is locked to the readings: walking back from the newest reading, each step lands on the reading closest to the expected mark if one is within two and a half minutes, otherwise on a virtual mark exactly five minutes on. That keeps the grid on sensor time through drift and crosses sensor gaps in even steps.

Each mark owns the time halfway to its neighbours, so every treatment and reading belongs to exactly one step. The pill lists all treatments in the block, then the readings, then any override or temp target at the mark, and stays the same across the whole block. Tap to select is unchanged.

Replaces #739, which addressed the same complaint about treatments repeating in the pill while scrubbing but did so by picking a single item and dropping the stacked pill.

New unit tests cover dense data, one-minute data, drift, stragglers, gaps and tiling.


codebymini (approved):
LGTM. Approve by code review and test

marionbarker (approved):
Approve by test on both Loop and Trio URLs.

PR #752, opened 2026-09-18
CI: Move remaining actions to Node 24 runtimes

bjorkert:
`actions/cache@v5` and `actions/github-script@v8` run on Node 24. These were the last two actions in the workflows still declaring Node 20, which GitHub now flags as deprecated on every run and forces onto Node 24 anyway. Every other action already targets Node 24.


marionbarker (approved):
Approved by code review

